PR #8, opened 2020-01-22
Fix metrics with no metadata

hunterwerlla:
- VSCode generator - fix generation with no `metadata`
- JetBrains - Update JetBrains generation tests to test for the same scenario

## License

### DIFF
--- a/telemetry/jetbrains/src/test/resources/testGeneratorInput.json
+++ b/telemetry/jetbrains/src/test/resources/testGeneratorInput.json
@@ -47,6 +47,11 @@
       "description": "called when invoking lambdas remotely",
       "unit": "None",
       "metadata": [{ "type": "lambdaRuntime", "required": false }, { "type": "inttype" }]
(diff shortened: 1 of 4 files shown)


telemetry/jetbrains/src/test/resources/testGeneratorInput.json
```diff
@@ -47,6 +47,11 @@
       "description": "called when invoking lambdas remotely",
       "unit": "None",
       "metadata": [{ "type": "lambdaRuntime", "required": false }, { "type": "inttype" }]
+    },
+    {
+      "name": "no_metadata",
```
justinmk3: why not give it a prefix/"namespace"? e.g. `lambda_no_metadata` ?

hunterwerlla: this is just test input, this puts it in the namespace "no" though

PR #10, opened 2020-01-23
Refactor IDE specific metrics to also be part of this repo

hunterwerlla:
- Refactor IDE specific metrics to be in jetbrainsMetrics/vscodeMetrics
- Update the generators to make use of these new files
- change `--input` to `--extraInput` for the Typescript generator

## License


telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/gradle/GenerateTelemetry.kt
```diff
@@ -21,7 +21,7 @@ open class GenerateTelemetry : DefaultTask() {
     fun generate() {
         println("Generating telemetry using packaged file and:\n ${inputFiles.map { it.absolutePath }.joinToString("\n")}")
         try {
-            TelemetryGenerator.generateTelemetryFromFiles(inputFiles, outputDirectory)
+            TelemetryGenerator.generateTelemetryFromFiles(inputFiles, outputFolder = outputDirectory)
```
bryceitoc9: Is this change necessary if you don't reorder the parameters in `generateTelemetryFromFiles`?

telemetry/jetbrains/build.gradle.kts
```diff
@@ -60,22 +59,21 @@ tasks {
...
             } catch (e: Exception) {
                 println("Exception while validating packaged schema, ${e.printStackTrace()}")
             }
         }
     }
     task(name = "copyTelemetryResources", type = Copy::class) {
-        from("..") {
-            include("*.json")
-        }
+        from("../telemetrySchema.json", "../definitions/commonDefinitions.json", "../definitions/jetbrainsDefinitions.json")
```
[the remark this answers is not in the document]
hunterwerlla: It does work but I want to specifically not pull in the vscode one (or Visual Studio one when that exists)

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/gradle/GenerateTelemetry.kt
hunterwerlla: No, but I think the parameter reorder makes more sense in my mind. I'll change these calls to all use named parameters.

PR #11, opened 2020-01-24
Migrate over more telemetry definitions

hunterwerlla:
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -85,7 +85,7 @@
         {
             "name": "lambda_invokelocal",
             "description": "Called when invoking lambdas locally (with SAM in most toolkits)",
-            "metadata": [{ "type": "runtime" }, { "type": "result" }, { "type": "debug"}]
+            "metadata": [{ "type": "runtime" }, { "type": "result" }, { "type": "debug" }]
         },
         {
             "name": "cloudformation_delete",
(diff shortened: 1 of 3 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -21,14 +21,15 @@
             "description": "A generic version metadata"
         },
         {
-            "name": "hasException",
-            "type": "boolean",
-            "description": "Determines if the operation had an exception or not"
+            "name": "reason",
+            "type": "string",
```
bryceitoc9: `reason` is typed in VS Code for `createSamApp`, should that be the case here too?: https://github.com/aws/aws-toolkit-vscode/blob/master/src/lambda/commands/createNewSamApp.ts#L58

hunterwerlla: No because this can be used for other things

bryceitoc9: Will those other things have similar reasons/can we just keep bulking up the type?

hunterwerlla: I'd rather not build a huge enum of reasons, this field is more of a get more information field (that could contain IDE specific information) than a field that we aggregate together across IDEs

bryceitoc9:
Please change `result` to `telemetryResult` as `result` is too common of a variable name to be reserved exclusively for telemetry.

telemetry/definitions/vscodeDefinitions.json
```diff
@@ -4,6 +4,57 @@
             "name": "vscode_activeregions",
```
bryceitoc9: Nit: `vscode_activeRegions`

hunterwerlla: Fixed

telemetry/vscode/src/generate.ts
```diff
@@ -5,11 +5,15 @@
 
 import { MetadataType, MetricMetadataType, MetricMetadata, Metric, MetricDefinitionRoot } from './parser'
 
-// converts snake_case to CamelCase. E.x. lambda_invoke => LambdaInvoke
+function toTitleCase(s: string): string {
+    return s.replace(s[0], s[0].toUpperCase())
```
bryceitoc9: Nit: Looks like the generally accepted way to do this is
```
return s[0].toUpperCase() + s.slice(1)
```

PR #17, opened 2020-02-05
Kotlin generation improvements

hunterwerlla:
## Description
- Move `from` into companion object
- Fix nullability of arguments and default value
- Do not attach metadata if it's null
- Log invalid `from` statements
## License

### DIFF
--- a/telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
+++ b/telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
@@ -13,6 +13,7 @@ import com.squareup.kotlinpoet.ParameterSpec
 import com.squareup.kotlinpoet.PropertySpec
 import org.slf4j.Logger
 import com.squareup.kotlinpoet.TypeSpec
+import com.squareup.kotlinpoet.asTypeName
 import java.io.File
 
 object TelemetryGenerator {
@@ -58,7 +59,7 @@ object TelemetryGenerator {
             .addFunction(
                 FunSpec.builder("from")
                     .returns(ClassName("", item.name.toTypeFormat()).copy(nullable = true))
-                    .addParameter("type", Any::class)
+                    .addParameter("type", Any::class.asTypeName().copy(nullable = true))
                     .addStatement("val result = values().firstOrNull { it.name == type.toString() }")
                     .beginControlFlow("if(result == null)")
                     .addStatement("LOG.warn(\"Invalid property \$type passed into %L\")", item.name.toTypeFormat())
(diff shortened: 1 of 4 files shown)


telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
```diff
@@ -40,21 +44,32 @@ object TelemetryGenerator {
...
-                .returns(ClassName("", item.name.toTypeFormat()).copy(nullable = true))
-                .addParameter("type", Any::class)
-                .addStatement("return values().firstOrNull { it.name == type.toString() }").build()
-        )
-        enum.addKdoc(item.description)
+
+        val companion = TypeSpec.companionObjectBuilder()
+            .addFunction(
+                FunSpec.builder("from")
+                    .returns(ClassName("", item.name.toTypeFormat()).copy(nullable = true))
```
kiiadi: To me `from` is effectively a constructor so it should never return null. Either you get back an object or it blows up.

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
```diff
@@ -40,21 +44,32 @@ object TelemetryGenerator {
...
+                    .addParameter("type", Any::class.asTypeName().copy(nullable = true))
```
kiiadi: Similarly, why can't this be non-nullable?

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
hunterwerlla: That was my initial thought, but I'm hesitant to throw as a result of telemetry. I can easily imagine that we make a change and forget to change the telemetry which would actually affect a workflow

hunterwerlla: Had a conversation offline, reverted this change and made it able to throw again

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
hunterwerlla: back to being non nullable

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
kiiadi: Not sure what the convo was, and I agree that generally we don't want Telemetry to break the general flow - but this should only be the case if there's a code-problem at build time - right? so it should be OK in this case.

PR #19, opened 2020-02-06
Convert all jetbrains telemetry

hunterwerlla:
## Description
- Add JetBrains metrics
- Add `createTime` to JetBrains telemetry calls as an optional parameter 
- make the output folder if it doesn't exist yet
- remove `record` from telemetry name (e.g. `S3Telemetry.recordUploadObject` -> `S3Telemetry.uploadObject`

## License


telemetry/definitions/clouddebugDefinitions.json
```diff
@@ -0,0 +1,87 @@
+{
+    "types": [
+        {
+            "name": "cloudDebugPlatform",
```
justinmk3: seems overwrought to have separate types for `cloudDebugPlatform`, `runtime`, `schemaLanguage`. Just like we don't have a separate type for `int` ranges, we could have 1 type that covers the general idea of a "platform".

This would make filtering easier.

telemetry/definitions/commonDefinitions.json
```diff
@@ -30,6 +30,11 @@
             "type": "string",
             "description": "The reason for a metric or exception depending on context"
         },
+        {
+            "name": "eventbridgeschema",
+            "type": "string",
+            "description": "The name of the EventBridge Schema used in the operation"
```
justinmk3: What is gained by having these hyper-specific types? Why not just use string?

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
```diff
@@ -86,7 +87,7 @@ object TelemetryParser {
         try {
             schema.validate(JSONObject(fileContents))
         } catch (e: Exception) {
-            System.err.println("Schema validation failed due to thrown exception $e")
+            System.err.println("Schema validation failed due to thrown exception $e\n on input: $fileContents")
```
justinmk3: if input is likely to be multiline it should start on its own line.

```suggestion
            System.err.println("Schema validation failed due to thrown exception $e\n on input:\n$fileContents")
```

telemetry/definitions/commonDefinitions.json
hunterwerlla: we could but the type also sets the name, it doesn't actually generate an `enum` or anything else for it in Kotlin. In typescript it generates comments of what each type is.

telemetry/definitions/clouddebugDefinitions.json
hunterwerlla: We have weird types currently for cloudDebugPlatform and schema language I don't think they really overlap. The type names here typically come from somewhere else so I'm hesitant to change them

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
hunterwerlla: Sure, will change

telemetry/definitions/clouddebugDefinitions.json
justinmk3: That's unfortunate, but if this is a legacy constraint then it is what it is :)

telemetry/definitions/commonDefinitions.json
```diff
@@ -30,6 +30,11 @@
             "type": "string",
             "description": "The reason for a metric or exception depending on context"
         },
+        {
+            "name": "eventbridgeschema",
```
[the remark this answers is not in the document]
hunterwerlla: Fixed

telemetry/definitions/commonDefinitions.json
```diff
@@ -72,10 +92,20 @@
...
+            "description": "Open the credentials file",
+            "metadata": [{ "type": "result" }]
+        },
         {
             "name": "aws_createCredentials",
             "description": "Create a new credentials file"
         },
+        {
+            "name": "aws_validateCredentials",
+            "description": "Validate set credentials",
```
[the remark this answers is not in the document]
hunterwerlla: Will add

telemetry/definitions/commonDefinitions.json
```diff
@@ -112,6 +142,11 @@
             "description": "called when creating lambdas remotely",
             "metadata": [{ "type": "runtime" }]
         },
+        {
+            "name": "lambda_editFunction",
+            "description": "called when creating lambdas remotely",
```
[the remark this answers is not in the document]
hunterwerlla: Fixed

telemetry/definitions/commonDefinitions.json
```diff
@@ -166,6 +202,61 @@
         {
             "name": "session_end",
             "description": "Called when stopping the IDE on a best effort basis"
+        },
+        {
+            "name": "s3_copyBucketName",
+            "description": "Copy the bucket name to the clipboard"
+        },
+        {
+            "name": "s3_copyPath",
```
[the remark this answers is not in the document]
hunterwerlla: Apparently not

PR #20, opened 2020-02-10
Add CodeBuild CI builds

hunterwerlla:
## Description
- Add a CI job for Node and Kotlin that will validate that changes will not break the repo

## License


buildspec/kotlinTests.yml
```diff
@@ -0,0 +1,18 @@
...
+    build:
+        commands:
+            - |
+                # TODO - iterate through the subdirectories to have one build script per runtime
+                cd telemetry/jetbrains
+                ./gradlew test
+
+artifacts:
+    files:
+        - '**/*'
```
justinmk3: wouldn't this be something like `build/dist/...` ? Seems wasteful to dump the entire tree into an artifacts store (and then will also cause big/noisy downloads for anyone who actually is interested in the artifacts)

hunterwerlla: This is a case of bad copy paste, for now we don't care about these artifacts, so removed

PR #23, opened 2020-02-17
Add remaining missing JetBrains definitions

hunterwerlla:
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -11,6 +11,11 @@
             "description": "The result of the operation"
         },
         {
+            "name": "resourceType",
+            "type": "string",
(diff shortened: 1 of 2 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -13,7 +13,7 @@
         {
             "name": "resourceType",
             "type": "string",
-            "description": "A generic name metadata"
+            "description": "The type of resource acted on, usually a service"
```
abrooksv: Would this be Lambda? Or lambdaFunction? Or what?

hunterwerlla: Fixed to be `serviceType` which captures what we actually want to do

PR #25, opened 2020-03-24
Add telemetry for CloudWatch Logs

hunterwerlla:
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -149,13 +149,13 @@
         },
         {
             "name": "cloudwatchlogs_openGroup",
-            "description": "Open the CloudWatch Logs group window",
-            "metadata": [{ "type": "result" }]
+            "description": "Open the CloudWatch Logs group window. ServiceType indicates that it was opened from a different service (like directly from an ECS container)",
+            "metadata": [{ "type": "result" }, {"type": "serviceType", "required": "false"}]
         },
         {
             "name": "cloudwatchlogs_openStream",
-            "description": "Open a CloudWatch Logs stream in the window",
-            "metadata": [{"type": "result"}]
+            "description": "Open a CloudWatch Logs stream in the window. ServiceType indicates that it was opened from a different service (like directly from an ECS container)",
+            "metadata": [{"type": "result"}, {"type": "serviceType", "required": "false"}]
         },
         {
             "name": "cloudwatchlogs_downloadStreamToFile",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -160,17 +160,17 @@
         {
             "name": "cloudwatchlogs_downloadStreamToFile",
             "unit": "Bytes",
-            "description": "Download a log stream to a file. Value indicates the final size of the formatted log stream.",
+            "description": "Download a stream to a file on disk. Value indicates the final size of the formatted stream.",
             "metadata": [{"type": "result"}]
         },
         {
             "name": "cloudwatchlogs_openStreamInEditor",
-            "description": "Download a log stream to open in editor.",
+            "description": "Download a stream to memory then open in an editor.",
             "metadata": [{"type": "result"}]
         },
         {
-            "name": "cloudwatchlogs_viewCurrentInEditor",
-            "description": "Export the currently open (possibly filtered) log stream to an editor",
+            "name": "cloudwatchlogs_viewCurrentMessagesInEditor",
+            "description": "Copy the currently open (possibly filtered) messages to an editor",
             "metadata": [{"type": "result"}]
         },
         {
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -12,9 +12,9 @@
         },
         {
             "name": "serviceType",
-            "allowedValues": ["ecs", "cloudformation", "lambda", "s3", "schemas"],
+            "allowedValues": ["ecs", "cloudformation", "lambda", "logs", "s3", "schemas"],
```
ckant: "logs" is the name in the SDK?

telemetry/definitions/commonDefinitions.json
```diff
@@ -142,6 +147,63 @@
...
+        {
+            "name": "cloudwatchlogs_openStream",
+            "description": "Open a CloudWatch Logs stream in the window",
+            "metadata": [{"type": "result"}]
+        },
+        {
+            "name": "cloudwatchlogs_downloadToFile",
+            "unit": "Bytes",
+            "description": "Download a log stream to a file. Value indicates the final size of the formatted log stream.",
+            "metadata": [{"type": "result"}]
```
ckant: Curious, you say the result here refers to the size. Do results of other metrics refer to whether the operation was a success?

If so, is there a way to indicate success/failure of the operation independently from the size here? i.e. the download succeeds/fails

telemetry/definitions/commonDefinitions.json
```diff
@@ -142,6 +147,63 @@
...
+            "description": "Open the CloudWatch Logs group window",
+            "metadata": [{ "type": "result" }]
+        },
+        {
+            "name": "cloudwatchlogs_openStream",
+            "description": "Open a CloudWatch Logs stream in the window",
+            "metadata": [{"type": "result"}]
+        },
+        {
+            "name": "cloudwatchlogs_downloadToFile",
```
ckant: `downloadStreamToFile` for consistency with other *stream operations?

telemetry/definitions/commonDefinitions.json
```diff
@@ -142,6 +147,63 @@
...
+        },
+        {
+            "name": "cloudwatchlogs_openStreamInEditor",
+            "unit": "Bytes",
+            "description": "Download a log stream to open in editor. Value indicates the final size of the formatted log stream.",
+            "metadata": [{"type": "result"}]
+        },
+        {
+            "name": "cloudwatchlogs_openCurrentInEditor",
+            "unit": "Bytes",
```
ckant: Why is unit bytes if it refers to number of lines?

telemetry/definitions/commonDefinitions.json
```diff
@@ -142,6 +147,63 @@
...
+            "metadata": [{"type": "result"}]
+        },
+        {
+            "name": "cloudwatchlogs_openStreamInEditor",
+            "unit": "Bytes",
+            "description": "Download a log stream to open in editor. Value indicates the final size of the formatted log stream.",
+            "metadata": [{"type": "result"}]
+        },
+        {
+            "name": "cloudwatchlogs_openCurrentInEditor",
```
ckant: What the difference between this and the above?

telemetry/definitions/commonDefinitions.json
```diff
@@ -142,6 +147,63 @@
...
+            "description": "Word wrap logs off/on",
+            "metadata": [{"type": "enabled"}]
+        },
+        {
+            "name": "cloudwatchlogs_tailLogs",
+            "description": "Tail logs off/on",
+            "metadata": [{"type": "enabled"}]
+        },
+        {
+            "name": "cloudwatchlogs_refreshLogGroup",
```
ckant: Does this just submit a count of presses or does it also include the total time taken?

telemetry/definitions/commonDefinitions.json
hunterwerlla: yep you can find these by doing a search for `SERVICE_NAME` in the Java SDK. Added that to the description as well

telemetry/definitions/commonDefinitions.json
hunterwerlla: Sure, changed

telemetry/definitions/commonDefinitions.json
hunterwerlla: Success comes from `result` but size comes from `Value` which along with `starttime` are two things that can always be set independent of everything else

telemetry/definitions/commonDefinitions.json
hunterwerlla: See the description, download a whole stream and open vs open current view

telemetry/definitions/commonDefinitions.json
hunterwerlla: Because I forgot to change it :P . I don't think number of lines will help us capture anything so just removed

telemetry/definitions/commonDefinitions.json
hunterwerlla: Just counts the number of presses, I'm skeptical of refresh time because it's more influenced by the service and what is being loaded so I don't think we'll get any data from it.

telemetry/definitions/commonDefinitions.json
ckant: Clarified: both deal with streams in memory. `Open` loads the entire stream (previously unopened) and shows it in the editor pane. `Current messages` shows the messages the user is looking at in the editor pane.

telemetry/definitions/commonDefinitions.json
ckant: That fair. But I think we should start capturing real performance metrics because it's the best way to see the bottlenecks that might only be visible under certain hardware / network situations.

PR #29, opened 2020-04-23
Add searchLogGroup metric

hunterwerlla:
Add a metric for searchLogGroup which is in PR currently

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -158,11 +158,6 @@
             "metadata": [{"type": "result"}, {"type": "serviceType", "required": false}]
         },
         {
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -196,6 +201,11 @@
             "description": "Called when a stream is searched",
             "metadata": [{"type": "result"}]
         },
+        {
+            "name": "cloudwatchlogs_searchGroup",
```
rli, line 205: this doesn't apply for speedsearch right? seems a bit too noisy if it does.

hunterwerlla: Nope, only for the search box that actually queries the service

PR #32, opened 2020-05-18
Adding telemetry definition for injecting credentials into 3rd party runs

kiiadi:


telemetry/definitions/commonDefinitions.json
```diff
@@ -122,6 +122,11 @@
             "name": "aws_createCredentials",
             "description": "Create a new credentials file"
         },
+        {
+            "name": "aws_injectCredentials",
+            "description": "Inject selected AWS credentials into a third-party run (e.g. RunConfiguration)",
+            "metadata": [{ "type": "result" }]
```
abrooksv, line 128: should we care about flavor? python vs java vs node? As an optional field?

kiiadi: I was wondering that too ... good call

kiiadi: so runtime?

abrooksv: sounds fine to be cocsistent

kiiadi: yeah - no - I've added a `runtimeString` value that is free text so that we don't have to introduce new logic when a new version of Node comes out or whatever.

PR #38, opened 2020-06-03
Changing Kotlin enums to be CamelCase instead of UPPER_CASE

kiiadi:
This more closely matches the Kotlin preferred code style.

### DIFF
--- a/telemetry/jetbrains/src/test/kotlin/software/aws/toolkits/telemetry/generator/GeneratorTest.kt
+++ b/telemetry/jetbrains/src/test/kotlin/software/aws/toolkits/telemetry/generator/GeneratorTest.kt
@@ -3,6 +3,7 @@
 
 package software.aws.toolkits.telemetry.generator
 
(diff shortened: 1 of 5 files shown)


telemetry/jetbrains/src/test/kotlin/software/aws/toolkits/telemetry/generator/GeneratorTest.kt
```diff
@@ -51,9 +52,17 @@ class GeneratorTest {
...
+
+        val expected = this.javaClass.getResourceAsStream(expectedOutputFile).use {
+            it.bufferedReader().readText()
+        }
+
+        try {
+            assertThat(actual).isEqualToIgnoringWhitespace(expected)
+        } catch (e: AssertionError) {
+            // This gives us better visualization of the differences in the IDE
+            throw ComparisonFailure(e.message!!, expected, actual)
```
hunterwerlla, line 65: Nice!

PR #39, opened 2020-06-03
Adding definitions for changing connections 

kiiadi:
It's useful to get metrics on regions and credential types being used - it will help prioritizing adding additional credential types across toolkits.


telemetry/definitions/commonDefinitions.json
```diff
@@ -116,7 +148,17 @@
         {
             "name": "aws_setCredentials",
             "description": "Select a credentials profile",
-            "metadata": [{ "type": "result" }]
+            "metadata": [{ "type": "result" }, { "type": "credentialType" }]
+        },
+        {
+            "name": "aws_setRegion",
+            "description": "A partition change occurred",
```
awschristou, line 155: Copy paste error?

telemetry/definitions/commonDefinitions.json
```diff
@@ -116,7 +148,17 @@
...
+            "metadata": [{ "type": "result" }, { "type": "credentialType" }]
```
awschristou, line 156: did you mean to use type regionId instead of credentialType? Similar for line 161

telemetry/definitions/commonDefinitions.json
kiiadi: good catch!

telemetry/definitions/commonDefinitions.json
kiiadi: yes - cos i'm a moron.

PR #40, opened 2020-06-03
Automate the maven release by closing the repo

hunterwerlla:
## License


telemetry/jetbrains/build.gradle.kts
```diff
@@ -143,3 +144,13 @@ signing {
         sign(publishing.publications["maven"])
     }
 }
+
+nexusStaging {
+    packageGroup = "software.aws"
+    serverUrl = "https://aws.oss.sonatype.org/service/local/"
```
kiiadi, line 150: Do we need to care about snapshot vs release?

hunterwerlla: No, since we don't build up multiple artifacts and release at the same time

PR #42, opened 2020-06-23
Add telemetry for RDS/Redshift

hunterwerlla:
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -350,7 +350,7 @@
         },
         {
             "name": "rds_getCredentials",
-            "description": "Called when getting IAM/SecretsManager credentials for an Redshift database. Value represents how long it takes in ms.",
+            "description": "Called when getting IAM/SecretsManager credentials for a RDS database. Value represents how long it takes in ms.",
             "unit": "Milliseconds",
             "metadata": [{ "type": "result" }, {"type": "databaseCredentials" }, {"type": "databaseEngine"}]
         },
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -12,7 +12,7 @@
         },
         {
             "name": "serviceType",
-            "allowedValues": ["ecs", "cloudformation", "lambda", "logs", "s3", "schemas", "stepfunctions"],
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -147,6 +147,17 @@
             "name": "enhancedHealthEnabled",
             "type": "boolean",
             "description": "Whether or not Elastic Beanstalk enhanced health reporting and monitoring is being used"
+        },
+        {
+            "name": "databaseEngine",
```
ckant, line 152: Reason for not using an enum?

telemetry/definitions/commonDefinitions.json
```diff
@@ -337,6 +348,28 @@
             "description": "Called when deleting a cloudformation stack",
             "metadata": [{ "type": "result" }]
         },
+        {
+            "name": "rds_getCredentials",
+            "description": "Called when getting IAM/SecretsManager credentials for an Redshift database. Value represents how long it takes in ms.",
```
ckant, line 353: typo "redshift"

hunterwerlla: Bad copy paste, fixed

telemetry/definitions/commonDefinitions.json
hunterwerlla: Since this value comes directly from the service, I don't think there is much value in adding an enum (because we just introduce the potential to have exceptions thrown as new things are added). SecretsManager auth in particular should work with DocumentDB/other engines that we do not support directly right now (with nodes or actions), and so it should be captured properly passing it through. Same with the serviceType, which we should maybe remove the enum list from.

hunterwerlla:
> I have no real opinion on generifying metric names
> 
> `{ "name": "getCredentials", "databaseType": "RDS|Redshift"}`
> vs
> `{ "name": "rds_getCredentials" }`, `{ "name": "redshift_getCredentials" }`
> 
> I think this is fine. Might make it slightly more annoying to compare apples to apples or use a generic metric publisher. But it also lets us treat each service specially instead of assuming they conform to the same interface.
> 
> Potentially more pain at impl / analysis to compare apples to apples vs. potentially more pain at impl / more confusion at analysis if the meanings behind the metrics diverge and we're comparing apples to oranges. Fine with the former.

Yeah I think this is a very rare case where they actually overlap entirely so I don't think it would make sense to add support for it. If we see more instances of this, it might make sense to allow.

PR #45, opened 2020-07-21
Add telemetry for snippet insertion

ckant:
## Types of changes

## Description

Add telemetry for snippet insertion

## Motivation and Context

This will track telemetry when the user selects and inserts a snippet into the current document.

Language is an enum to ensure values with consistent names and casing.

## Related Issue(s)

aws/aws-toolkit-vscode#1173

## Checklist

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -159,6 +159,17 @@
             "type": "string",
             "allowedValues": ["IAM", "SecretsManager"],
             "description": "How the database credentials are being retrieved"
+        },
+        {
+            "name": "snippetPrefix",
+            "type": "string",
+            "description": "The prefix of a snippet"
+        },
+        {
+            "name": "snippetLanguage",
+            "allowedValues": ["javascript", "typescript"],
+            "type": "string",
+            "description": "The language of a snippet"
         }
     ],
     "metrics": [
@@ -476,6 +487,11 @@
             "name": "toolkit_init",
             "description": "The Toolkit has completed initialization",
             "metadata": [{ "type": "duration" }]
+        },
+        {
+            "name": "snippet_insert",
+            "description": "Insert a snippet",
+            "metadata": [{ "type": "snippetPrefix" }, { "type": "snippetLanguage" }]
         }
     ]
 }


justinmk3:
> Language is an enum to ensure values with consistent names and casing.

Are we duplicating vscode API language ids? Can we use those instead somehow.

telemetry/definitions/commonDefinitions.json
```diff
@@ -159,6 +159,17 @@
...
             "allowedValues": ["IAM", "SecretsManager"],
             "description": "How the database credentials are being retrieved"
+        },
+        {
+            "name": "snippetPrefix",
+            "type": "string",
+            "description": "The prefix of a snippet"
+        },
+        {
+            "name": "snippetLanguage",
```
hunterwerlla, line 169: Can we use runtimeString instead of making more languages? (runtimeString can be renamed if you like because it's only used in a thing that is unused currently)

telemetry/definitions/commonDefinitions.json
```diff
@@ -159,6 +159,17 @@
             "type": "string",
             "allowedValues": ["IAM", "SecretsManager"],
             "description": "How the database credentials are being retrieved"
+        },
+        {
+            "name": "snippetPrefix",
```
hunterwerlla, line 164: Why do we need this? why not just name?

telemetry/definitions/commonDefinitions.json
abrooksv: That wouldn't tell us TS vs js.

hunterwerlla: runtimeString is a generic runtime name that is currently only used with `aws_injectCredentials` (which is in draft PR in JB)

ckant: I'm thinking it's not really a runtime. It corresponds to the language of the code (`javascript` vs. `nodejs`, `java` vs `java8` or `OpenJDK8`, `csharp` vs `dotnet`, etc.).

The exact correspondence I'm modelling is called [`language identifier`](https://code.visualstudio.com/docs/languages/identifiers#_known-language-identifiers) in vscode. Jetbrains calls these "contexts" under which the live templates are made available.

Also it seems useful to make an enum here so that we don't end up with `java` and `Java`.

telemetry/definitions/commonDefinitions.json
ckant: Yeah I considered name. No problem either way. Name makes more sense if it's trying to be generic with something like Live Templates. Prefix makes more sense if we use the vscode terminology.

awschristou:
We will hold off on snippet related telemetry until we revisit that functionality at a later time. Closing PR.

PR #46, opened 2020-07-23
C# Telemetry type separation and code rearrangements

awschristou:
## Description

This is the first in a series of changes to get C# related Toolkit Telemetry code in a place where it can be packaged for consumption by the Toolkit.

* The solution was code generator centric. The files have been arranged, and renamed, to better reflect their place among additional C# telemetry related code that will be added to the repo.
* namespaces have been adjusted to have an `Amazon.AwsToolkit.Telemetry` prefix, to better organize current and future C# telemetry related code that will be added to the repo.
* the autogenerated "telemetry events" code no longer references Telemetry SDK related datatypes. A `AwsToolkit.Telemetry.Events` project has been started, which contains the types referenced by the autogenerated code. This allows the Toolkit to use the telemetry emitting code without having a dependency on the aws sdk (AWSSDK.Core).
* Future tasks will add support for the generator to produce code for supplemental telemetry definitions. The generator now outputs a partial class, to prepare for this work.
* Tests have been added that exercise the autogenerated telemetry events code. To accomplish this, a `PreBuild` step within [AwsToolkit.Telemetry.Events.Tests.csproj](.\telemetry\csharp\AwsToolkit.Telemetry.Events.Tests\AwsToolkit.Telemetry.Events.Tests.csproj) builds and runs the generator, emitting a file into the test project
* CI has been adjusted to use GitHub Actions, until we can make a CodeBuild job that supports both .NET Core and .NET Framework. The generator used to be the only C# code in the repo, and it is .NET Core, but the newly added `AwsToolkit.Telemetry.Events` must be .NET Framework (4.6) for the Toolkit to use it.

Upcoming tasks (future PRs)
* generator changes, to produce additional code files off of repo-centric telemetry definitions
* NuGet packaging and publishing (these will be last, as we develop a clearer picture of how the generator will be integrated into the Toolkit build scripts)

## License


telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Core/MetricDatumExtensionMethodsTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            Assert.Equal("false", _sut.Metadata[Key]);
+        }
+
+        [Fact]
+        public void AddMetadata_BlankEntries()
+        {
+            _sut.AddMetadata(Key, (object) null);
+            Assert.False(_sut.Metadata.ContainsKey(Key));
+
+            _sut.AddMetadata(Key, (string) null);
```
kiiadi, line 39: whitespace testing?

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/GeneratedCodeTests.cs
```diff
@@ -0,0 +1,108 @@
...
+                {
+                    _recordedMetrics = metrics;
+                });
+        }
+
+        /// <summary>
+        /// RecordLambdaInvokeRemote was chosen as a sample call that uses a telemetry "enum" type (Runtime)
+        /// </summary>
+        [Fact]
+        public void RecordLambdaInvokeRemote()
```
kiiadi, line 29: Do you see this as a test to prove out telemetry as a whole (eg. if these pass we assume all generated code will pass)? Or would we need to change these as we add new metrics?

buildspec/csharpTests.yml
```diff
@@ -6,12 +6,12 @@ phases:
...
-            - dotnet build -c Release ./telemetry/csharp/ToolkitTelemetryGenerator.sln
-            - dotnet test -c Release ./telemetry/csharp/ToolkitTelemetryGenerator.sln --logger trx --results-directory ./.testresults
-
-reports:
-    DotnetTestExamples:
-        file-format: VisualStudioTrx
-        files:
-            - '**/*'
-        base-directory: './.testresults'
+            - echo "C# CodeBuild jobs removed until an image with .NET Framework and .NET Core can be created (using GitHub Actions in the meantime)"
```
abrooksv, line 9: We do this in JB:

```
phases:
  install:
    runtime-versions:
      java: corretto11
      dotnet: 3.1

    commands:
      - apt update
      - apt install -y mono-complete
```

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Core/MetricDatumExtensionMethodsTests.cs
awschristou: Nice catch, I've added a test that passes in spaces.

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/GeneratedCodeTests.cs
awschristou: In an ideal world, we'd auto generate code that tests each of the generated methods.

Because all of the RecordXxx call implementations are predictable, I think its reasonable to pick a sampling of calls that exercise the different handling scenarios/types. I don't see us adding tests for every new metric that is added.

buildspec/csharpTests.yml
awschristou: Thank you, I didn't know about this. I'll explore it further outside of this PR.

PR #51, opened 2020-08-20
Add passive flag to metrics

rli:
## License


telemetry/service/service-model.json
```diff
@@ -1,205 +1,236 @@
 {
-    "version": "2.0",
-    "metadata": {
-        "apiVersion": "2017-07-25",
-        "endpointPrefix": "client-telemetry",
```
awschristou, line 5: This changed from `client-telemetry` to `ToolkitTelemetry` -- did that change the generated C# code?

telemetry/csharp/client-generator-inputs/telemetry-2017-07-25.normal.json
```diff
@@ -0,0 +1,205 @@
+{
```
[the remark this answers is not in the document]
rli: removed

telemetry/service/service-model.json
rli: it changes the "RegionEndpointServiceName" property, but since we're not in the endpoints.json, i don't think it matters, but reverted just in case it does

PR #53, opened 2020-08-28
Add API Gateway metrics

hunterwerlla:
Running prettier formatted more than I thought it would, but the changes are:

- adding `apigateway` to the service ID list
- The addition of `httpMethod`
- The `apigateway_*`

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -12,7 +12,19 @@
         },
         {
             "name": "serviceType",
-            "allowedValues": ["apigateway", "ecs", "cloudformation", "lambda", "rds", "redshift", "logs", "s3", "schemas", "stepfunctions", "sqs"],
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -163,6 +180,31 @@
...
+                { "type": "debug" }
+            ]
+        },
+        {
+            "name": "apigateway_invokeRemote",
+            "description": "Calling a remote API Gateway",
+            "metadata": [{ "type": "result" }, { "type": "httpMethod", "required": false }]
+        },
+        {
+            "name": "apigateway_startLocalServer",
```
awschristou, line 204: Is this a passive event?

hunterwerlla: Nope, it requires user input (the action does not exist today yet)

PR #55, opened 2020-09-02
Add metrics for SQS

hunterwerlla:
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -546,7 +546,8 @@
         },
         {
             "name": "sqs_openQueue",
-            "description": "Open an SQS queue. Initially opens to the send message pane or poll messags pane."
+            "description": "Open an SQS queue. Initially opens to the send message pane or poll messags pane.",
+            "metadata": [{ "type": "sqsQueueType" }]
         },
         {
             "name": "sqs_createQueue",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -23,7 +23,7 @@
                 "s3",
                 "schemas",
                 "stepfunctions",
-				"sns",
+                "sns",
                 "sqs"
             ],
             "type": "string",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -177,6 +178,11 @@
             "type": "string",
             "allowedValues": ["IAM", "SecretsManager"],
             "description": "How the database credentials are being retrieved"
+        },
+        {
+            "name": "sqsQueueType",
+            "allowedValues": ["standard", "fifo"],
+            "description": "The type of an SQS Queue (either standard or fifo)"
```
justinmk3, line 185: nit: if we have valid values enumerated programmatically, better to leave them out of descriptions (or auto-generate the description). Small thing, but adds up to a maintenance cost over time

telemetry/definitions/commonDefinitions.json
```diff
@@ -537,6 +543,57 @@
...
+        {
+            "name": "sqs_deleteMessages",
+            "description": "Delete one or more messages from an SQS queue. Value indicates the number of messages that we tried to delete.",
+            "unit": "Count",
+            "metadata": [{ "type": "result" }, { "type": "sqsQueueType" }]
+        },
+        {
+            "name": "sqs_copyMessage",
+            "description": "Copy a message to the clipboard",
+            "metadata": [{ "type": "sqsQueueType" }]
```
justinmk3, line 576: we might instead want a generic `copyToClipboard` metric with perhaps a parameter that indicates the component that initiated it. 

Though having telemetry on copy/paste in general seems a bit much, almost like tracking how many keys the user pressed...

justinmk3 (approved):

telemetry/definitions/commonDefinitions.json
hunterwerlla: Yeah, maybe we should just not have this. It doesn't really tell us anything, removed.

telemetry/definitions/commonDefinitions.json
hunterwerlla: Sure, removed from the description

PR #66, opened 2020-09-24
Add S3 CopyUrl

abrooksv:
## License


awschristou (approved):

hunterwerlla (approved):

justinmk3:
Similar thing was proposed in https://github.com/aws/aws-toolkit-common/pull/55#discussion_r482354523 but I wonder if we really want metrics on "copy to clipboard".

hunterwerlla:
For presigned URLs we will be making a service call so I think it's reasonable in this case

PR #71, opened 2020-09-28
Add unknown to all enums

abrooksv:
* Make from method return unknown if none match instead of throwing error
* Remove un-needed toString when type is already a string
* Add another suppression to make it ignore "could be private"

## License


telemetry/jetbrains/src/test/resources/testResultOutput
```diff
@@ -32,12 +31,14 @@ enum class Result(
 
     DashSeparated("dash-separated"),
 
-    UnderscoreSeparated("underscore_separated");
+    UnderscoreSeparated("underscore_separated"),
+
+    Unknown("unknown");
```
kiiadi, line 36: Assume we should follow this pattern with all languages?

abrooksv: Probs? I havent looked at the others yet.

In JetBrains we failed when we added new lambda runtimes and forgot telemetry so I ant to fall back to unknown in that case

abrooksv: TS seems to be `foo|bar` so it will fail to compile so dont think we need it there

abrooksv: C# does not generate a static from method, so probably not needed there either?

kiiadi: what was the failure in JetBrains? was it a build time thing? cos if so - maybe it should fail so we see it....

telemetry/jetbrains/src/test/resources/testResultOutput
```diff
@@ -32,12 +31,14 @@ enum class Result(
...
 
     override fun toString(): String = value
 
     companion object {
-        fun from(type: Any): Result = values().first { it.value == type.toString() }
```
hunterwerlla, line 40: Why does this no longer take any `Any`?

telemetry/jetbrains/src/test/resources/testResultOutput
abrooksv: Nah, we converted the Lambda Runtime enum value to this version though the value `([Telemetry]Runtime.from([LambdaRuntime]runtime.toString)`

telemetry/jetbrains/src/test/resources/testResultOutput
abrooksv: Why would it take an Any? 
We forced the value to be a String, so when would we not want it to take a string in the from? 

Implicitly relying on a toString to match the value of this seems more dirty than making the call site be aware that it owns the conversion

hunterwerlla (approved):

PR #72, opened 2020-09-30
Add new telemetry definition to understand when credentials are loaded

kiiadi:
Adding a new telemetry definition around credential loading. In order to optimize certain workflows it's useful to understand how often credentials are loaded and how many credentials are present.

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -122,6 +122,11 @@
             "description": "True if turned on, false if turned off"
         },
         {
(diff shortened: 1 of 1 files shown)


abrooksv:
Why do we need credential type and credential type id?

kiiadi:
> Why do we need credential type and credential type id?

The way we load credentials is by factory - this will allow us to disambiguate between other factory types in the future.

abrooksv:
Ah it's the factory type id. Makes sense.

abrooksv (approved):

PR #77, opened 2020-10-02
Deployment related metrics

awschristou:
## Description

This change adds some metrics related to the deploy features in AWS Toolkit for Visual Studio.

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -200,6 +206,11 @@
             "type": "boolean",
             "description": "Whether or not it was a presigned request"
         },
+        {
+            "name": "platform",
+            "type": "string",
+            "description": "Language-specific identification"
```
bryceitoc9, line 212: What does this represent? Doesn't seem language-specific if it's paired with a runtime?

telemetry/definitions/commonDefinitions.json
```diff
@@ -517,6 +558,17 @@
             "description": "Called when updating a Lambda function's code outside the context of a SAM template",
             "metadata": [{ "type": "runtime", "required": false }, { "type": "result" }]
         },
+        {
+            "name": "lambda_deploy",
```
bryceitoc9, line 562: What does this metric represent? Right now, we have two separate metrics for uploads:
* JB uses `lambda_editFunction` (JB can update both function configuration and code)
* VS Code will use `lambda_updateFunctionCode` 

Neither have a scratch-create Lambda. Is that what this represents?

telemetry/definitions/commonDefinitions.json
awschristou: This is a little more open ended, but you'd get language specific values, for example:
C#: `v4.6.1`, `netcoreapp2.1`, `netcoreapp3.1`
node: `nodejs10.x`, `nodejs12.x`

There may be some overlap with Lambda runtimes, but this isn't strictly tied to lambda usage. Runtimes like `provided` (for Custom Runtimes) are opaque, so this also provides a chance to get additional detail. (this is how we would have captured .NET Core 3.0 usage)

telemetry/definitions/commonDefinitions.json
awschristou: VS has configuration editing capabilities, which lambda_editFunction works for too.

JB and VS Code lean into SAM for their publishing operations, which is the sam_deploy metric. But they also allow a more one-off "update this code" operation. In VS, there is no SAM CLI related experience, rather its publishing operation is Lambda specific.

bryceitoc9: OK, so if I understand this correctly, VS will emit both `editFunction` and `deploy`, with `deploy` occurring on both `Lambda.createFunction` and `Lambda.updateFunctionCode` calls (with `initialDeploy` true/false, respectively) and `editFunction` happening on `Lambda.updateFunctionConfiguration`?

If that's the case, we should standardize on those metrics sooner rather than later...could you cut an issue to make that change in JB? I can make the change to VS Code no sweat since Lambda upload is still in staging.

telemetry/definitions/commonDefinitions.json
hunterwerlla: Can that be added to the description?

telemetry/definitions/commonDefinitions.json
```diff
@@ -446,6 +467,26 @@
...
+            "metadata": [{ "type": "result" }, { "type": "regionId" }, { "type": "ecsLaunchType" }]
+        },
+        {
+            "name": "ecs_deployTask",
+            "description": "Called when deploying a task to an ECS cluster",
+            "metadata": [
+                { "type": "result" },
+                { "type": "regionId" },
+                { "type": "ecsLaunchType" },
+                { "type": "scheduled" }
```
hunterwerlla, line 487: what does scheduled mean?

telemetry/definitions/commonDefinitions.json
```diff
@@ -214,6 +225,11 @@
             "name": "insightsQueryStringType",
             "allowedValues": ["insights", "searchTerm"],
             "description": "User-specified search string type while starting an insights query"
+        },
+        {
+            "name": "ecsLaunchType",
+            "allowedValues": ["ec2", "fargate"],
+            "description": "Type of ECS Task"
```
hunterwerlla, line 232: "Compute environment for ECS task"?

telemetry/definitions/commonDefinitions.json
```diff
@@ -402,6 +418,11 @@
             "unit": "Milliseconds",
             "metadata": [{ "type": "result" }]
         },
+        {
+            "name": "cloudformation_deploy",
+            "description": "Called when deploying a CloudFormation template",
+            "metadata": [{ "type": "result" }, { "type": "regionId" }, { "type": "initialDeploy" }]
```
hunterwerlla, line 424: when did `initalDeploy` get into these metrics? What question does this answer?

telemetry/definitions/commonDefinitions.json
awschristou: Done

telemetry/definitions/commonDefinitions.json
awschristou: ECS tasks can be [scheduled](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/scheduling_tasks.html) as part of the deployment process.

(Line 134's description: "Whether or not the referenced item is scheduled (true) or not (false)")

telemetry/definitions/commonDefinitions.json
awschristou: Revised to "Infrastructure type used by ECS tasks and services" based on https://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_types.html

telemetry/definitions/commonDefinitions.json
awschristou: `initialDeploy` was introduced with beanstalk deploy, and signals whether a deploy is targeting a new destination (like a create), or an existing destination (like an update).

hunterwerlla: I get that, but why do we care? It's the same API call and UI regardless?

telemetry/definitions/commonDefinitions.json
justinmk3: I would suggest we don't need the granularity of "edit properties" vs "edit code", but otherwise if we are going to separate them, agree with @bryceitoc9 that the semantics should be consistent across IDEs.

telemetry/definitions/commonDefinitions.json
awschristou: This provides insight into whether we are seeing more successes/failures around creates vs updates, as well as an understanding around whether Toolkits are used more for creates or updates.

awschristou: > It's the same API call and UI

These are dangerous assumptions to make across the Toolkits. In VS, CloudFormation template deployment uses the same entrypoints to deploy a new or updated stack, and the deploy wizard has some variations based on user inputs. 

The API calls used are not as relevant to the metric around the act of deployment.  [CreateStack](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html) or [UpdateStack](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_UpdateStack.html) are used as needed.

telemetry/definitions/commonDefinitions.json
kiiadi: the nice thing about having separate metric for edit properties vs edit code - is that they are actually different workflows - if one is more popular than the other, and is missing as an option in other tools - it may drive us to change our functionality.

telemetry/definitions/commonDefinitions.json
hunterwerlla: I think this should be optional, or not have it at all. What questions does this answer?

telemetry/definitions/commonDefinitions.json
bryceitoc9: Regardless, we have a potential redundancy: are `updateFunctionCode` and `deploy` with `initialDeploy: false` equivalent? If so, we should center on one (which would likely be `deploy` with `initialDeploy: false`) and we should do the following things:
* change VS Code to use `deploy`
* break apart Jetbrains' `editFunction` to also emit `deploy` if necessary (this seems like something we should do either way)

telemetry/definitions/commonDefinitions.json
awschristou: When you deploy in VS, deploying ECS tasks and deploying scheduled ECS tasks are top level items, so it indicates how much each workflow is used.

awschristou: Tasks and Scheduled Tasks are top level Cluster entities, just like Services, so the "scheduled" field should be required.

image

hunterwerlla: If it's a different top level thing, should it not be its own metric?

awschristou: It is using the same construct (Tasks), so I wouldn't think so.

hunterwerlla (approved):

PR #85, opened 2020-10-10
Fix case where generated supplemental metrics code is unable to reference common telemetry types

awschristou:
## Description

The generator was set up with the assumption that generated code would reside in the namespace `Amazon.AwsToolkit.Telemetry.Events`. Now that we generate NuGet packages, we know that the generated code lives in Amazon.AwsToolkit.Telemetry.Events.Generated`. 

This change fixes a bug in generated code where supplemental telemetry definitions reference types from the common telemetry definitions (example: `Result`). The generated supplemental code expected those types to live in `Amazon.AwsToolkit.Telemetry.Events`, causing compilation errors.

## License

### DIFF
--- a/telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
+++ b/telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
@@ -7,7 +7,6 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-using Amazon.AwsToolkit.Telemetry.Events;
 using Amazon.AwsToolkit.Telemetry.Events.Core;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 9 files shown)


telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
```diff
@@ -16,7 +15,7 @@
 /// This file is generated from https://github.com/aws/aws-toolkit-common/tree/master/telemetry
 /// --------------------------------------------------------------------------------
 
-namespace Amazon.AwsToolkit.Telemetry.Events.Tests.Generated
+namespace Amazon.AwsToolkit.Telemetry.Events.Generated
```
awschristou, line 18: You don't need to review the changes to this file. It is produced by the generator and used in some tests.
This line here shows how the bug wasn't caught in the tests, because the supplemental generated and this generated file were produced to the same namespace, meaning the supplemental code had direct access to types (like Result) from this file.

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/SupplementalCode.cs
```diff
@@ -7,8 +7,8 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
```
awschristou, line 8: You don't need to review the changes to this file. It is produced by the generator and used in some tests.

shruti0085 (approved):

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/SupplementalCode.cs
justinmk3: it's useful to see the diff so I'm glad it's in source control

justinmk3 (approved):

PR #88, opened 2020-10-28
Allow overriding existing telemetry types with the extra input files

hunterwerlla:
Allow overriding existing telemetry types with the extra input files

## License

### DIFF
--- a/telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
+++ b/telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
@@ -86,6 +86,12 @@ data class MetadataSchema(
 object TelemetryParser {
     private val MAPPER = jacksonObjectMapper()
 
+    /**
+     * Read in default definitions and extra definitions files, and deserialize into an object
+     *
+     * Definitions from `paths` will override definitions from defaultResourceFiles if their names
+     * are the same. This allows testing updates to existing definitions from within each IDE project.
+     */
     fun parseFiles(
         defaultResourcesFiles: List<String>,
         paths: List<File> = listOf()
@@ -101,8 +107,6 @@ object TelemetryParser {
                 it.metrics.plus(it2.metrics)
             )
         }.let {
-            // Allow read in files to overwrite default definitions. First one wins, so the extra
-            // files are read before the default resources
             TelemetryDefinition(
                 it.types.distinctBy{ t -> t.name},
                 it.metrics.distinctBy { m -> m.name }
--- a/telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
+++ b/telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
@@ -87,8 +87,8 @@ object TelemetryParser {
     private val MAPPER = jacksonObjectMapper()
 
     fun parseFiles(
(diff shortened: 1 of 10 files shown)


awschristou (commented):
Is there documentation for how this works?
I see Kotlin and TS support, what about C#?

hunterwerlla:
> Is there documentation for how this works?
> I see Kotlin and TS support, what about C#?

Added documentation. The C# generator works differently, so this will not currently work with it as written.

kiiadi (commented):
Do you think we should be explicit in the intention that a definition will override a core one (and bomb if it doesn't)? In this case what if we added `lambda_updation`, thinking we were overriding the central `lambda_update` definition with some additional attributes - but we've actually inadvertently introduced a new definition.

telemetry/vscode/src/generateTelemetry.ts
```diff
@@ -50,14 +51,20 @@ function formatOutput(output: string) {
...
-    const input: MetricDefinitionRoot = args.inputFiles.map(validateInput).reduce(
+    const rawDefinitions: MetricDefinitionRoot = args.inputFiles.map(validateInput).reduce(
         (item: MetricDefinitionRoot, input: MetricDefinitionRoot) => {
             item.types!.push(...(input.types ?? []))
             item.metrics.push(...input.metrics)
             return item
         },
         { types: [], metrics: [] }
     )
+    // Allow read in files to overwrite default definitions. First one wins, so the extra
```
awschristou, line 62: It would be nice to see a test for TS similar to `test/kotlin/software/aws/toolkits/telemetry/generator/GeneratorTest.kt` - there doesn't appear to be any validation around this change.

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
```diff
@@ -100,6 +100,13 @@ object TelemetryParser {
                 it.types.plus(it2.types),
                 it.metrics.plus(it2.metrics)
             )
+        }.let {
+            // Allow read in files to overwrite default definitions. First one wins, so the extra
```
awschristou, line 104: This should probably be a part of the function comment, documenting the specific behavior of the generator with respect to args.

awschristou (approved):
I would prefer to see override types/metrics made explicit...
* allows validation of intent to override 
* allows identification of overrides metrics versus product-specific metrics (easier to locate and promote metrics to central definitions later on)
* can prevent/validate against "accidental" overrides by product-specific metrics

This change is serviceable though.

telemetry/vscode/src/generateTelemetry.ts
hunterwerlla: That is outside of the scope of this PR, there are no tests for the Typescript generator (we have an item in the backlog for it)

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryParser.kt
hunterwerlla: Added

hunterwerlla:
> Do you think we should be explicit in the intention that a definition will override a core one (and bomb if it doesn't)? In this case what if we added `lambda_updation`, thinking we were overriding the central `lambda_update` definition with some additional attributes - but we've actually inadvertently introduced a new definition.

I'm conflicted because I'm not sure where we would stick it. In the schema as an optional property? In that case it's useless in the non overrided files which is weird? Maybe as an explicit override file? I'll file an issue to revisit this

PR #90, opened 2020-11-12
Add tests for the VS Code telemetry generator

hunterwerlla:
- Add tests based off of the JB tests
- Refactor the code to test better, use the prettier api to format

## License


telemetry/vscode/src/generate.ts
```diff
@@ -127,8 +139,49 @@ export function generateTelemetry(telemetryJson: MetricDefinitionRoot): string {
...
+        })
+        .reduce(
+            (item: MetricDefinitionRoot, input: MetricDefinitionRoot) => {
+                item.types!.push(...(input.types ?? []))
+                item.metrics.push(...input.metrics)
+                return item
+            },
+            { types: [], metrics: [] }
+        )
+    // Allow read in files to overwrite default definitions. First one wins, so the extra
```
justinmk3, line 165: does "read in" mean "discovered" 
```suggestion
    // Allow discovered files to overwrite default definitions. First one wins, so the extra
```

telemetry/vscode/src/generateTelemetry.ts
```diff
@@ -3,13 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
```
justinmk3, line 3: nit (code structure): seems backwards that the entry point is `generateTelemetry.ts` which actually invokes `generate.ts`. And/or the files could be combined, don't see a strong reason to separate them.

telemetry/vscode/tst/generatorTest.ts
```diff
@@ -0,0 +1,40 @@
+/*!
```
justinmk3, line 1: nit (naming): why "tst" instead of "test"?

- in the [jetbrains case](https://github.com/aws/aws-toolkit-common/tree/73e05585f323cea9f17d19b9d3f88f3209d4b7cd/telemetry/jetbrains/src/test) the test files live in `src/test/`
- in https://github.com/aws/aws-toolkit-vscode tests live in `src/test/`

As a habit it is good to avoid entropy if there is not a strong reason for the entropy.

telemetry/vscode/tst/generatorTest.ts
```diff
@@ -0,0 +1,40 @@
...
+    beforeAll(() => {
+        tempDir = tmpdir()
+    })
+
+    test('Generate fails when validation fails', async () => {
+        try {
+            await testGenerator(['resources/invalidInput.json'], '/invalid/file/path' )
+        } catch (e) {
+            expect(e).not.toBeNull
+        }
```
justinmk3, line 21: if the generator doesn't throw then the test passes. 

```suggestion
            return
        }
        assert.fail()
```

telemetry/vscode/tst/generatorTest.ts
```diff
@@ -0,0 +1,40 @@
...
+    })
+
+    test('Generate overrides', async () => {
+        await testGenerator(['resources/testOverrideInput.json', 'resources/testResultInput.json'], 'resources/generatorOverrideOutput')
+    })
+
+    async function testGenerator(inputFiles: string[], expectedOutputFile: string) {
+        const output = `${tempDir}/output`
+        await generate({ inputFiles: inputFiles.map(item => `${__dirname}/${item}`), outputFile: output })
+        // TODO remove spaces and line returns so that it matches more generally? pull in a better matching lib?
```
justinmk3, line 35: https://www.npmjs.com/package/ast-compare looks pretty lean (few dependencies)

telemetry/vscode/tst/parserTest.ts
```diff
@@ -0,0 +1,47 @@
+/*!
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { validateInput } from '../src/parser'
```
justinmk3, line 6: I wonder if `parser` should be renamed to `validator`.

justinmk3 (approved):

telemetry/vscode/tst/generatorTest.ts
hunterwerlla: It's different because I'm bad at naming things consistently, it was not a conscious change. fixed

telemetry/vscode/tst/generatorTest.ts
hunterwerlla: Forgot this, added

telemetry/vscode/tst/generatorTest.ts
hunterwerlla: It works fine as is, but maybe that would be good in the future

telemetry/vscode/tst/parserTest.ts
hunterwerlla: Yeah it does both so I'm not sure

telemetry/vscode/src/generateTelemetry.ts
hunterwerlla: generateTelemetry.ts is the script that runs everything, it's what is actually called by the toolkit and does argument parsing as well. I think it makes sense to have that be separate.

telemetry/vscode/src/generate.ts
hunterwerlla: There's no discovery happening? The files that are specified explicitly to be part of the generate

PR #106, opened 2021-04-06
Document awsAccount and awsRegion as metrics properties; add them to C# generator

awschristou:
## Description

This change documents `awsAccount` and `awsRegion` as properties available to all metrics.

This change also updates the C# telemetry generator so that these properties are available to the VS Toolkit. I also added some constant values ("n/a", "not-set", "invalid") to have some consistency with the JetBrains implementation.

## License

### DIFF
--- a/telemetry/telemetryformat.md
+++ b/telemetry/telemetryformat.md
@@ -55,6 +55,10 @@ _Additionally_ two additional global arguments that can be appended to any gener
 createTime?: Date
 // Value based on unit and call type
 value?: number
+// The AWS Region associated with a metric
+string? awsAccount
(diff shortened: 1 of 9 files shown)


hunterwerlla (approved):

telemetry/telemetryformat.md
```diff
@@ -102,6 +106,10 @@ interface LambdaRemoteinvoke {
     createTime?: Date
     // Value based on unit and call type,
     value?: number
+    // The AWS Region associated with a metric
+    awsAccount?: string
```
gsweetwood: typo: should be 'AWS Account'

PR #107, opened 2021-04-07
CodeCommit metrics

awschristou:
## Description

This change defines CodeCommit related metrics.

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -447,6 +447,27 @@
...
+            "metadata": [{ "type": "result" }]
+        },
+        {
+            "name": "codecommit_createRepo",
+            "description": "A repo is created in CodeCommit",
+            "metadata": [{ "type": "result" }]
+        },
+        {
+            "name": "codecommit_setCredentialsAutomatically",
+            "description": "A connection is automatically established to CodeCommit to perform actions on repos",
```
hunterwerlla, line 462: What is the difference between this and the one below it? Doesn't this just fit into active vs passive?

awschristou: It is possible for the Toolkit to auto-connect on startup, and it is possible for a user to explicitly connect. They could be the same metric, where the passive flag is variable. I wasn't sure if we encourage adjusting the passive flag at runtime or if the preference is to "fix" the passive value when it is defined.

hunterwerlla: There are other instances of metrics where we want them to be active sometimes and passive other times. I think it makes sense as 1 metrics with active or passive set based on what happens

awschristou: Ok excellent, honestly I opened the PR with this same question, so I'm happy we discussed this.

hunterwerlla (approved):

telemetry/definitions/commonDefinitions.json
justinmk3: Does the typescript telemetry client support setting `passive` on any metric?

awschristou: I don't know if it does, I know the C# one doesn't, I'm in the process of making a change to the C# generator to allow calling code to set passive on any metric.

PR #108, opened 2021-04-09
Update C# telemetry generator to allow client code to adjust if a metric is passive

awschristou:
## Description

The VS Toolkit hasn't had a need to designate certain metrics as either passive or active. #107 identifies a scenario (`codecommit_setCredentials`) where we have both active and passive activities raising this event.

This change updates the C# generator to allow client code to adjust the `Passive` state of a metric. To accomplish this, a field "Passive" has been added to the base class for all metric classes (`BaseTelemetryEvent`). This class is initialized within every metric class constructor.

## License

### DIFF
--- a/telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
+++ b/telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
@@ -44,7 +44,7 @@ namespace Test
                 var datum = new MetricDatum();
                 datum.MetricName = "emit_someIntMs";
                 datum.Unit = Unit.Milliseconds;
-                datum.Passive = false;
+                datum.Passive = payload.Passive;
                 if (payload.Value.HasValue)
                 {
                     datum.Value = payload.Value.Value;
@@ -86,7 +86,7 @@ namespace Test
                 var datum = new MetricDatum();
                 datum.MetricName = "emit_noUnit";
                 datum.Unit = Unit.None;
-                datum.Passive = false;
+                datum.Passive = payload.Passive;
                 if (payload.Value.HasValue)
                 {
                     datum.Value = payload.Value.Value;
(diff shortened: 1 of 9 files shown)


telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
```diff
@@ -322,11 +322,21 @@ namespace Test
...
+        public EmitSomeIntMs()
+        {
+            this.Passive = false;
+        }
     }
     
     /// Emits something without a unit defined
     public sealed class EmitNoUnit : BaseTelemetryEvent
     {
+
```
hunterwerlla, line 335: nit: why is there an empty line at the start but not at the bottom? looks funky

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
```diff
@@ -6147,11 +6332,21 @@ public sealed class AwsLoadCredentials : BaseTelemetryEvent
...
+        }
     }
     
     /// Create a new credentials file
     public sealed class AwsCreateCredentials : BaseTelemetryEvent
     {
+        
+        public AwsCreateCredentials()
+        {
+            this.Passive = false;
```
hunterwerlla, line 6348: nit: why is this not an argument to `BaseTelemetryEvent`?

hunterwerlla (approved):

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
awschristou: Its an artifact of the code generation library.

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
awschristou: Currently the code-generation code is very brute force. It would have been more complicated to have the code generator produce code that passes this to a base class constructor, but is a good candidate for a future improvement.

PR #110, opened 2021-04-26
Add credentialSourceId to set credentials metric

awschristou:
## Description

This change adds an optional field to the "set credentials" metric, which allows us to determine where the credentials came from. This will give us a better indication of usage patterns across different credentials sources.

For example, in the VS Toolkit, the credentials profile can currently from from the encrypted SDK Credentials store, or the Shared Credentials files.

`windowsCredentialStore` has been removed from `credentialType` because it is a container for a credentials definition rather than a type.

## License


rli:
doesn't `credentialType` already encompass the range of sources?

awschristou:
> doesn't `credentialType` already encompass the range of sources?

No. I notice that credentialType contains `windowsCredentialStore` -- this should probably be removed. The SDK Store can contain most (if not all) of the same credentials types that a shared credentials file can, which is the motivation for this new (optional) axis.

rli (approved):

PR #111, opened 2021-04-26
Do not maintain a list of serviceIds since they come from the sdks anyway

hunterwerlla:
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -17,29 +17,8 @@
         },
         {
             "name": "serviceType",
-            "allowedValues": [
-                "apigateway",
-                "beanstalk",
-                "cloudformation",
-                "cloudfront",
-                "dynamodb",
-                "ec2",
-                "ecr",
-                "ecs",
-                "iam",
-                "lambda",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -17,29 +17,8 @@
...
-                "s3",
-                "schemas",
-                "sns",
-                "sqs",
-                "stepfunctions",
-                "vpc"
-            ],
             "type": "string",
-            "description": "The name of the AWS service acted on. These values come from the AWS SDK. To find them in the JAVA SDK search for SERVICE_NAME in each service client."
+            "description": "The name of the AWS service acted on. These values come from the AWS SDK. To find them in the JAVA SDK search for SERVICE_NAME in each service client, or look for serviceId in metadata in the service2.json"
```
awschristou, line 21: Are they the same values across SDKs? 
Does it make sense to comment about Java specifically?

hunterwerlla: Yep, that's why I updated the description to include the source of serviceid

awschristou (approved):

PR #112, opened 2021-04-26
Add source data for ECR deployments

rli:
## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -531,7 +536,11 @@
         {
             "name": "ecr_deployImage",
             "description": "Called when deploying an image to ECR",
-            "metadata": [{ "type": "result" }, { "type": "regionId" }]
+            "metadata": [
+                { "type": "result" },
+                { "type": "regionId" },
+                { "type": "ecrDeploySource", "required": false }
```
awschristou, line 542: Is the intention to note when you are deploying **in** to ECR by Tag, or could this represent deploying **from** ECR to somewhere else by Tag? If it is the latter, we should consider a separate metric.

rli: this is local -> ECR, which is what i interpreted this metric to be currently representing

awschristou (approved):

PR #114, opened 2021-04-28
Credentials and Region standardization

awschristou:
## Description

Some changes have been made to standardize credentials and region metrics:
* credentialSourceId: changed to an enumeration, indicating where credentials were sourced from (for example shared credentials, the .NET sdk credentials store, or ec2). Previously this was a freeform string that was used inconsistently.
* credentialType: removed `envVars`, this was an eager definition entry that has been moved to credentialSourceId
* regionId: this field has been removed. The awsRegion property available to all metrics should be used in its place.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -142,11 +142,6 @@
             ]
         },
         {
-            "name": "regionId",
-            "type": "string",
-            "description": "The ID of the region that was selected"
-        },
-        {
             "name": "partitionId",
             "type": "string",
             "description": "The ID of the partition that was selected"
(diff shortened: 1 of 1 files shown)


kiiadi:
is this backwards compatible?

awschristou:
> is this backwards compatible?

These changes came out of a discussion with @justinmk3 and @abrooksv . They aren't backwards compatible in the sense that the values emitted to the credentialSourceId will change as we get a chance to update the Toolkits, and we'll stop using regionId. These changes will align our metrics on a going forward basis. We haven't been setting up visualizations in the affected spaces so far.

justinmk3 (approved):

kiiadi:
I meant more for integration into the toolkits when we do generation - if you've engaged with @justinmk3 and @abrooksv then I see no issue.

PR #117, opened 2021-05-04
Add the ability to override passive on JetBrains metrics

hunterwerlla:
- Also upgrade to Gradle 7 because it required no code changes

## License


telemetry/jetbrains/gradlew
```diff
@@ -159,14 +175,9 @@ save () {
     for i do printf %s\\n "$i" | sed "s/'/'\\\\''/g;1s/^/'/;\$s/\$/' \\\\/" ; done
     echo " "
 }
-APP_ARGS=$(save "$@")
+APP_ARGS=`save "$@"`
```
justinmk3, line 178: This is a stylistic regression, backticks are harder to read, can't be nested, and not part of POSIX, so not a "best practice"

hunterwerlla: This is gradle editing gradle's file on update, this file is not manually touched

rli (approved):

PR #191, opened 2021-08-13
Add credential source to 'validateCredentials'

JadenSimon:
Adds optional metadata to `validateCredentials` to check from where the credentials came from. 

## License


justinmk3 (approved):
cc @awschristou

PR #217, opened 2021-09-24
Collapse dynamo fetch definitions

kiiadi:


telemetry/definitions/commonDefinitions.json
```diff
@@ -239,6 +239,16 @@
...
             "allowedValues": ["dockerfile", "tag"],
             "description": "The source content specified in the ECR deployment request"
+        },
+        {
+            "name": "dydbFetchType",
+            "allowedValues": ["scan", "query"],
+            "description": "The type of fetch being performed"
+        },
+        {
+            "name": "dydbIndexType",
```
justinmk3, line 249: if we can't generalize this to something service-agnostic, I would suggest at least we don't introduce different service prefixes (dydb vs dynamodb)
```suggestion
            "name": "dynamodbIndexType",
```

it is unfortunate that we didn't use the `{prefix}_{name}` (with underscore) convention as in the metric names. Probably shouldn't start using underscores in type names I guess.

justinmk3 (approved):
+100 for collapsing definitions where possible. This is "synthesis", which encodes knowledge passively.

awschristou (approved):

PR #248, opened 2021-11-17
Modify s3_deleteObject and remove s3_deleteObjects

rbbarad:
Modify s3_deleteObject and remove s3_deleteObjects
- adds successCount and failedCount metadata to bulk s3 operations
- migrating away from plural metric names for bulk s3 operations (does not affect existing code)

This is an approach to simplify emitting bulk operation metrics - if we like this approach, we can do this to more metrics in the future.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -887,7 +887,7 @@
         },
         {
             "name": "s3_deleteObject",
-            "description": "Delete a single S3 object",
+            "description": "Delete S3 object(s)",
             "metadata": [
                 { "type": "result" },
                 { "type": "successCount", "required": false},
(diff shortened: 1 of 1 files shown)


KodyFintak (approved):

JadenSimon (approved):

KodyFintak (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -277,6 +277,16 @@
...
             "allowedValues": ["session-manager-plugin", "dotnet-lambda-deploy", "dotnet-deploy-cli"]
+        },
+        {
+            "name": "successCount",
+            "type": "double",
+            "description": "The number of successful operations"
+        },
+        {
+            "name": "failedCount",
+            "type": "double",
```
shruti0085, line 288: is there a reason why we want this to be double? since it's a count I believe it should be fine being an int.

KodyFintak (approved):

telemetry/definitions/commonDefinitions.json
rbbarad: initial rationale was to prevent any chances of overflow. Changed to int since that's not much of an issue

PR #268, opened 2022-01-14
Add metric for AWS modify credentials

rbbarad:
## Types of changes

## Description

Add metric for aws_modifyCredentials

## Motivation and Context

Allows us to track errors during modifications to the users' credentials profile

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -53,9 +53,10 @@
             "description": "A generic version metadata"
         },
         {
-            "name": "operation",
+            "name": "credentialModification",
             "type": "string",
-            "description": "A generic operation metadata"
+            "allowedValues": ["Add", "Edit", "Delete"],
+            "description": "The type of modification performed on the credentials"
         },
         {
             "name": "source",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -52,6 +52,16 @@
             "type": "string",
             "description": "A generic version metadata"
         },
+        {
+            "name": "operation",
```
awschristou, line 56: operation sounds like something we might want consistency between Toolkits. If that is the case, lets make this a definition that has allowed values for Add/Edit/Delete, and give it a less general name.

rbbarad: Would 'modificationType' be a better alternative here?

awschristou (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -52,6 +52,17 @@
...
             "description": "A generic version metadata"
         },
+        {
+            "name": "credentialModification",
+            "type": "string",
+            "allowedValues": ["Add", "Edit", "Delete"],
+            "description": "The type of modification performed on the credentials"
+        },
+        {
+            "name": "source",
```
JadenSimon, line 62: This will probably become a 'primitive' of telemetry at some point (e.g. not metadata)
Keeping track of an originator is extremely common in web browser telemetry as well as related protocols (think HTTP/HTTPS).
Just a little brain-dump.

PR #273, opened 2022-01-25
Define Create Project metrics for Lambda and CloudFormation

rbbarad:
## Types of changes

## Description

Define Create Project metrics for Lambda and CloudFormation.

## Motivation and Context

Allows us to track the types of projects being created (languages and templates used).

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -566,6 +571,11 @@
             "unit": "Milliseconds",
             "metadata": [{ "type": "result" }]
         },
+        {
+            "name": "cloudformation_createProject",
+            "description": "Called when creating a CloudFormation project",
+            "metadata": [{ "type": "templateName" }]
```
awschristou, line 577: We should emit result for both metrics, because the process could fail (or be cancelled)

rbbarad: ah good catch!

awschristou (approved):

PR #280, opened 2022-02-01
Fix check for `undefined` values for VS Code generator

JadenSimon:
The `if` statement should be checking for `undefined`, not falsy values.

## License


awschristou (approved):
Is this change isolated, or is this something you'd want to check for in this code's lint rules (for example)

JadenSimon:
The generated code itself had the flaw :/
But yeah probably should lint for these imprecise checks, even if it means a bit more verbosity when you actually do want to check for any non-falsy value.

PR #290, opened 2022-02-28
Add datapoint for deeplinks

rli:
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1284,7 +1284,8 @@
                 {
                     "type": "result"
                 }
-            ]
+            ],
+            "passive": true
         }
     ]
 }
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1273,6 +1273,18 @@
                 }
             ],
             "passive": true
+        },
+        {
+            "name": "deeplink_open",
+            "description": "User requested that a resource be opened in the browser using the deeplink service",
+            "metadata": [
+                {
+                    "type": "source"
+                },
+                {
+                    "type": "result"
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1273,6 +1273,18 @@
                 }
             ],
             "passive": true
+        },
+        {
+            "name": "deeplink_open",
+            "description": "User requested that a resource be opened in the browser using the deeplink service",
+            "metadata": [
+                {
+                    "type": "source"
```
awschristou, line 1282: I assume "Type" represents where within the Toolkit that a link resided.
Any value in adding an optional "Service" field, to determine which services are more frequently used with deeplinks?

telemetry/definitions/commonDefinitions.json
```diff
@@ -1273,6 +1273,18 @@
                 }
             ],
             "passive": true
+        },
+        {
+            "name": "deeplink_open",
```
awschristou, line 1278: Should this be passive by default?

rli: links are only opened when users click on them, so i think it's appropriate to keep this as an active metric

telemetry/definitions/commonDefinitions.json
rli: the deeplinks team can give us that data, so i'd rather not have every toolkit try to figure out how to parse the service segment out of an ARN

telemetry/definitions/commonDefinitions.json
rli: sync'd offline; will make passive by default in case that use-case is more prevalent in the future

awschristou (approved):

PR #296, opened 2022-03-17
Add optional handler to adjust metadata prior to logging metrics

awschristou:
## Description

This change updates the generated metrics code to provide an optional callback that allows the caller to transform the metric properties prior to sending them to the server. This could be used for example to chop a long value into smaller segments.

This change was co-authored with @floralph 

## License


floralph (approved):
Pair programmed with chrstou@ on this

PR #311, opened 2022-05-11
telemetry: add "file_editAwsFile" metric

justinmk3:
## Problem

We want to have insight into whether customers are using "authoring" features for AWS filetypes (CFN, SAM, credentials, ...) in the AWS Toolkits.

## Solution

Add `file_editAwsFile `.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -13,6 +13,24 @@
             "description": "The current state of the App Runner service"
         },
         {
+            "name": "awsFiletype",
+            "type": "string",
+            "allowedValues": [
+                "awsCredentials",
+                "cloudformation",
+                "cloudformationSam",
+                "codebuildBuildspec",
+                "ecsTask",
+                "eventbridgeSchema",
+                "iamPolicy",
+                "stepfunctionsAsl",
+                "smithyModel",
+                "ssmDocument",
+                "other"
+            ],
+            "description": "AWS filetype kind"
+        },
+        {
             "name": "duration",
             "type": "double",
             "description": "The duration of the operation in milliseconds"
@@ -46,6 +64,11 @@
             "name": "name",
             "type": "string",
             "description": "A generic name metadata"
+        },
+        {
+            "name": "filenameExt",
+            "type": "string",
+            "description": "Filename extension (examples: .txt, .yml, .yaml, .asl.yaml, ...), or empty string if the filename does not contain dot (.) between two chars."
         },
         {
             "name": "attempts",
@@ -783,6 +806,15 @@
             "name": "feedback_result",
             "description": "Called while submitting in-IDE feedback",
             "metadata": [{ "type": "result" }]
+        },
+        {
+            "name": "file_editAwsFile",
+            "description": "Use authoring features such as autocompletion, syntax checking, and highlighting, for AWS filetypes (CFN, SAM, etc.). Emit this _once_ per file-editing session for a given file. Ideally this is emitted only if authoring features are used, rather than merely opening or touching a file.",
+            "metadata": [
+                { "type": "result" },
+                { "type": "awsFiletype", "required": true},
+                { "type": "filenameExt", "required": false}
+            ]
         },
         {
             "name": "iam_openRole",


telemetry/definitions/commonDefinitions.json
```diff
@@ -1000,6 +1022,16 @@
                 { "type": "failedCount", "required": false}
             ]
         },
+        {
+            "name": "aws_filetypes_edit",
+            "description": "Use authoring features such as autocompletion, syntax checking, and highlighting, for AWS filetypes (CFN, SAM, etc.)",
+            "metadata": [
+                { "type": "result" },
+                { "type": "enabled", "required": true},
```
justinmk3, line 1030: If Toolkits have a way to check that authoring features are disabled, this would be `false` if user disables such features.

telemetry/definitions/commonDefinitions.json
```diff
@@ -47,6 +61,11 @@
             "type": "string",
             "description": "A generic name metadata"
         },
+        {
+            "name": "filenameExt",
+            "type": "string",
+            "description": "Filename extension (examples: .txt, .yml, .yaml, .asl.yaml, ...), or empty string if the filename does not contain dot (.) between two chars"
```
justinmk3, line 67: Do we also want a generic "file opened" metric to see what kinds of files people are editing? This would only track file extension, for privacy (and we could set a hard limit of 8 chars).

telemetry/definitions/commonDefinitions.json
```diff
@@ -312,7 +338,7 @@
             "name": "toolId",
             "type": "string",
             "description": "The tool being installed",
-            "allowedValues": ["session-manager-plugin", "dotnet-lambda-deploy", "dotnet-deploy-cli"]
+            "allowedValues": ["session-manager-plugin", "dotnet-lambda-deploy", "dotnet-deploy-cli", "aws-cli"]
```
awschristou, line 341: Sanity check: did you intend to add this change? It is not in the PR Description (same with lines 121-123)

telemetry/definitions/commonDefinitions.json
```diff
@@ -1000,6 +1026,16 @@
...
             ]
         },
+        {
+            "name": "aws_filetypes_edit",
+            "description": "Use authoring features such as autocompletion, syntax checking, and highlighting, for AWS filetypes (CFN, SAM, etc.). Emit this _once_ per file-editing session for a given file. Ideally this is emitted only if authoring features are used, rather than merely opening or touching a file.",
+            "metadata": [
+                { "type": "result" },
+                { "type": "enabled", "required": true},
+                { "type": "awsFiletype", "required": true},
+                { "type": "filenameExt", "required": true}
```
awschristou, line 1036: What does extension provide that isn't implied by filetype? Would it be reasonable to make this field optional?

telemetry/definitions/commonDefinitions.json
awschristou: Do we have at least one Toolkit that is able to disable these features today? If not, this seems like something we could add in the future once it is applicable.

telemetry/definitions/commonDefinitions.json
awschristou: I don't think the Toolkits should emit something so open ended. As a user this would feel too invasive, because it doesn't appear to have relevance to the AWS extension. IMO it would be preferable to measure something specific that we intend to take action on.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1000,6 +1026,16 @@
                 { "type": "failedCount", "required": false}
             ]
         },
+        {
+            "name": "aws_filetypes_edit",
```
awschristou, line 1030: This is a departure from the current metric naming conventions. I don't have strong preference here, but its worth discussing before accepting this.

Style-wise, I don't know if the second underscore will break any assumptions around metric names.

Scanning the current metrics, it looks like this would be called something like `aws_editAwsFile`. I am guessing you want to scope the metric to "file related operations", which is handy if we add more file related metrics in the future. Today, the first portion of metric names roughly correlates to scope, what if we named it `file_editAwsFile`?

telemetry/definitions/commonDefinitions.json
justinmk3: > this seems like something we could add in the future once it is applicable.

It would need to be an optional field then (in the future). In general I lean towards making fields that we know about "required" so that everything is explicit (and thus more discoverable/noticeable) in client calls.

Happy to revert this though.

telemetry/definitions/commonDefinitions.json
justinmk3: I'm avoiding optional fields because explicit calls avoid entropy, help discovery, and serve as a sanity-check (caller must make a decision).

telemetry/definitions/commonDefinitions.json
awschristou: I agree with the philosophy. This field seems like YAGNI to me, but I don't have a strong preference to leave or remove it.

telemetry/definitions/commonDefinitions.json
JadenSimon: > This is a departure from the current metric naming conventions. I don't have strong preference here, but its worth discussing before accepting this.

Yeah for these definitions I'd prefer not deviating from convention just yet (it makes it harder to do widespread refactoring). `file_editAwsFile` can be thought of as a metric for the 'file' subsystem which I think that's the intent?

I'm okay with a different format so long as we know how it should be interpreted.

telemetry/definitions/commonDefinitions.json
JadenSimon: Same PR is fine IMO (though to be fair I do this myself way more than I should). Expanding the allowed values is non-breaking.

telemetry/definitions/commonDefinitions.json
JadenSimon: I'd suggest removing it if it's not implemented. It's okay if it is made optional later on. The absence of metadata itself is a differentiating factor among metrics. Forcing callers to assign a value means we lose that additional state. In other words, metrics emitted prior to the feature being implemented would be hard to differentiate from metrics after the feature is implemented despite conveying different information.

It's interesting how many of the best-practices in writing software don't translate well for data science. We actively avoid making something optional because that's potentially one more possible state that needs to be handled. Yet optional fields encode more information that's useful for analysis. Metrics may superficially look the same, but are they semantically equivalent?

Requiring metadata essentially implies that it is known ahead of time, which is implementation-dependent. One workaround to this problem is to make the tri-state explicit: don't use a boolean, use something like `'yes' | 'no' | 'unavailable'` instead. It's pretty gross but it's one way to differentiate between the presence and absence of a functionality.

telemetry/definitions/commonDefinitions.json
rli: consolas is doing this so let's model it

telemetry/definitions/commonDefinitions.json
```diff
@@ -12,6 +12,24 @@
             "allowedValues": ["CREATE_FAILED", "RUNNING", "DELETED", "DELETE_FAILED", "PAUSED", "OPERATION_IN_PROGRESS"],
             "description": "The current state of the App Runner service"
         },
+        {
+            "name": "awsFiletype",
```
JadenSimon, line 16: I'm being pedantic but 'awsLanguage' might be a better name? More generic too.

JadenSimon (approved):
Approving on the condition that we change the metric name to `awsfiletypes_edit`

telemetry/definitions/commonDefinitions.json
justinmk3: https://github.com/aws/aws-toolkit-common/pull/318

telemetry/definitions/commonDefinitions.json
justinmk3: removed

telemetry/definitions/commonDefinitions.json
justinmk3: > As a user this would feel too invasive, because it doesn't appear to have relevance to the AWS extension

It is only the extension, and only for filenames we support.

awschristou (approved):

telemetry/definitions/commonDefinitions.json
justinmk3: Many of these are just plain JSON and thus more of a format than language. I was also looking for something that is more distinctive as a topic--I sense that we want something _less_ generic in this case.

PR #317, opened 2022-05-18
Add dotnet6 lambda runtime

shruti0085:
Adds .NET 6 as a Lambda runtime.
Also updates generated code with latest definitions.

## License

### DIFF
--- a/telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
+++ b/telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
@@ -6981,6 +6981,55 @@ namespace Amazon.AwsToolkit.Telemetry.Events.Generated
         }
         
         /// Records Telemetry Event:
+        /// Create an SNS Topic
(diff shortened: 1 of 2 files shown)


rbbarad (approved):

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs
```diff
@@ -8240,6 +8289,9 @@ public struct Runtime
         /// dotnet5.0
         public static readonly Runtime Dotnet50 = new Runtime("dotnet5.0");
         
+        /// dotnet6.0
+        public static readonly Runtime Dotnet60 = new Runtime("dotnet6.0");
```
normj, line 8293: The Lambda runtime enum value is `dotnet6`. Not sure if this code is meant match exactly that.

shruti0085: In my understanding we do seem to that. Thanks for the callout!

PR #324, opened 2022-06-03
telemetry: lift definitions out of aws-toolkit-vscode

justinmk3:
- ~~BREAKING CHANGE: rename `lambdaArchitecture` to `architecture`~~
- Lift vscode metrics defs from https://github.com/aws/aws-toolkit-vscode/pull/2102
- Backwards-compatible change: add redundant `lambdaArchitecture` field to metrics with `architecture` field

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -66,9 +66,14 @@
             "description": "The Lambda Package type of the function"
         },
         {
+            "name": "architecture",
+            "allowedValues": ["x86_64", "arm64"],
+            "description": "DEPRECATED. Use lambdaArchitecture instead."
+        },
+        {
             "name": "lambdaArchitecture",
             "allowedValues": ["x86_64", "arm64"],
-            "description": "The Lambda Architecture of the function"
+            "description": "Lambda architecture identifier"
         },
         {
             "name": "serviceType",
@@ -379,7 +384,9 @@
                 { "type": "runtime", "required": false },
                 { "type": "httpMethod", "required": false },
                 { "type": "result" },
-                { "type": "debug" }
+                { "type": "debug" },
+                { "type": "architecture", "required": false },
+                { "type": "lambdaArchitecture", "required": false }
             ]
         },
         {
@@ -959,7 +966,9 @@
                 { "type": "version", "required": false },
                 { "type": "lambdaPackageType" },
                 { "type": "result" },
-                { "type": "debug" }
+                { "type": "debug" },
+                { "type": "architecture", "required": false },
+                { "type": "lambdaArchitecture", "required": false }
             ]
         },
         {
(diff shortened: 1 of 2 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -51,7 +51,7 @@
             "description": "The Lambda Package type of the function"
         },
         {
-            "name": "lambdaArchitecture",
+            "name": "architecture",
```
justinmk3, line 54: Renaming this because it is identical with the `architecture` type, which is used in several other metrics. Should be easy to update this in all toolkits @rli @awschristou

awschristou: `lambdaArchitecture` was defined in-product Sept 2021 (the same time as the VS Code Toolkit change). It was migrated over in Feb 2022. There was an internal discussion about this before adding lambdaArchitecture into the definitions.

justinmk3: `architecture` has been used in 

- `apigateway_invokeLocal`
- `lambda_invokeLocal`
- `sam_init`
- `sam_attachDebugger` (which will likely be removed)

, which are now being lifted into common (previously was local to vscode).

We will have outdated metrics whichever choice we make. Or I could add a redundant field (`lambdaArchitecture`) to those metrics.

I'll do that since there is concern about the current PR.

JadenSimon: Personally, I'd let both types exist but not add redundant fields. This is the only true non-breaking change from a consumer's perspective. Otherwise you will have fragmented the data-set.

The fundamental problem here is that our telemetry design associates data types 1:1 with field names. This goes against most modern IDLs (think Smithy, OpenAPI, WADL) as well as pretty much every strongly-typed programming language that supports interfaces.

Names of fields describe additional context applicable to the associated data type. But right now we cannot express that. The information is bound to the data type itself.

JadenSimon: > The datastore doesn't care about types. So we should use consistent names.

Our datastore actually _does_ care about types, we just automatically make everything a string :)
But yeah consistent naming is important. It's just that our current design gets in the way of sharing types with the same semantics but not the same context.

> An alternative is to have both names and drop the deprecated one a year from now. That is what this PR aims to do.

Yup I see what your intent was now. For some reason I thought we were going to have Toolkits emit one or the other, not both. If both are emitted, then that works.

telemetry/definitions/commonDefinitions.json
```diff
@@ -982,7 +991,9 @@
                 { "type": "version", "required": false },
                 { "type": "lambdaPackageType", "required": false },
                 { "type": "reason", "required": false },
-                { "type": "eventBridgeSchema", "required": false }
+                { "type": "eventBridgeSchema", "required": false },
+                { "type": "architecture", "required": false },
```
JadenSimon, line 995: Should we just keep `architecture` confined to the VSC Toolkit? It's somewhat counter to the goal of this PR, though it may be less confusing for other Toolkits.

JadenSimon: Or even just give up on `architecture` entirely as I think it's just specific to VSC. We can fix the data on the backend as-needed.

justinmk3: That makes sense. Because it's vscode-only, none of our main dashboards are using this field.

awschristou (approved):

PR #325, opened 2022-06-03
Define generalized CloudWatch Logs related events

awschristou:
## Description

Define CloudWatch Logs metrics for a set of operations. Some of these metrics are more generalized versions of existing definitions, which help to simplify our definitions in the long run. These definitions were designed and reviewed internally with @JadenSimon and @shruti0085.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -31,9 +31,24 @@
             "description": "AWS filetype kind"
         },
         {
+            "name": "cloudWatchLogsPresentation",
+            "allowedValues": ["ui", "text"],
+            "description": "Presentation mode used in a CloudWatch Logs operation"
+        },
+        {
+            "name": "cloudWatchResourceType",
+            "allowedValues": ["logGroup", "logGroups", "logStream"],
+            "description": "CloudWatch Logs entity"
+        },
+        {
             "name": "duration",
             "type": "double",
             "description": "The duration of the operation in milliseconds"
+        },
+        {
+            "name": "filterBy",
+            "allowedValues": ["prefix", "time"],
+            "description": "Type of filter applied"
         },
         {
             "name": "result",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -30,11 +30,26 @@
             ],
             "description": "AWS filetype kind"
         },
+        {
+            "name": "cloudWatchLogsPresentation",
+            "allowedValues": ["ui", "text"],
+            "description": "Presentation mode used in a CloudWatch Logs operation"
```
shruti0085, line 36: I believe this only applies to when we are viewing any cloudwatch logs resource. What do you think about updating the description to indicate this?

shruti0085 (approved):
Looks good to me. Deferring to @JadenSimon for any additional comments

telemetry/definitions/commonDefinitions.json
awschristou: I generalized that notion to "presentation mode", in case we wind up with a future metric that can leverage this type.

telemetry/definitions/commonDefinitions.json
```diff
@@ -30,11 +30,26 @@
...
+        },
+        {
+            "name": "cloudWatchResourceType",
+            "allowedValues": ["logGroup", "logGroups", "logStream"],
```
JadenSimon, line 40: Looks like we're treating a _collection_ of a resource as a concept in its own right. I've thought about this a lot and I think it's appropriate, however, we should think of a better name than just `logGroups`. Too easily confused with the singular `logGroup`. It may not be clear that we're refreshing the data provided by a `list` call rather than simply refreshing multiple log groups.

`logGroupCollection` is verbose though it may be more obvious? Maybe we don't make changes for this PR, but let's see if we can figure out a standard way to describe this idea.

telemetry/definitions/commonDefinitions.json
```diff
@@ -567,20 +582,55 @@
...
+                { "type": "cloudWatchResourceType" }
+            ]
+        },
+        {
+            "name": "cloudwatchlogs_open",
+            "description": "Open a CloudWatch Logs entity. ServiceType indicates where the request came from (example: while viewing an ECS container)",
+            "metadata": [
+                { "type": "result" },
+                { "type": "cloudWatchResourceType" },
+                { "type": "serviceType", "required": false },
```
JadenSimon, line 599: Is `serviceType` optional because it's not always known prior to executing the user gesture, or is it optional because its absence implies that the user initiated the action directly on a CW resource?

These kinds of subtleties will become increasingly important over time. Allowing a field to be optional effectively _increases_ the API surface area and is a backward compatible change. So in theory, all new metric fields should be required unless it's clear about what the optionality means.

We've done a good job with this in the rest of the PR so I'm not really concerned about this particular metric. Just something to think about.

JadenSimon (approved):

telemetry/definitions/commonDefinitions.json
awschristou: When I was reading the design doc to make this change, I initially thought it was a typo - "logGroups and logGroup?" - I like your suggestion to reduce confusion. Will update.

telemetry/definitions/commonDefinitions.json
awschristou: Here I'm using the service type field to represent "where did someone initiate a gesture to view a cloudwatch logs entity from?" - many of those gestures will be embedded within service related views, but the primary entry point in the AWS Explorer will not be service specific. 

Some alternative ideas:
- use an open-ended string field (which is inherently optional), but accept that different clients may place different values in the field
- define an all new enum type like "cloudwatchGestureSource", which we can add specific gesture points to
- (other suggestions?)

JadenSimon: I've been leaning towards a single, open-ended `source` field for things like this though nothing is really set-in-stone here. It's less useful for comparing across clients, but really this level of detail is already difficult to compare anyway as we're quickly approaching into UI.

CWL is a special case because it does relate to many, many services. So having a dedicated field just to label which service the gesture relates to could work in combination with a generic `source` that differentiates between UI.

Like I want to know _where_ the user started the action and for _what_ purpose.

awschristou: Ok, how about we leave serviceType (and keep it optional), and add a open ended `source` field. That way we can represent the two axis (and account for gestures from AWS Explorer). I think we would be more interested in comparing service type across clients than the actual source, so this gives us a nice level of flexibility.

JadenSimon (approved):

PR #328, opened 2022-06-07
feat(vscode/telemetry): export interfaces

justinmk3:
## Problem
The "kwargs" parameter-map interfaces are not exported, so clients often end up redundantly defining their own similar interfaces. PR https://github.com/aws/aws-toolkit-common/pull/2 does not indicate that this was intentional.

## Solution
Export the interfaces.

Also update the package-lock.json lockefileVersion.

## License

### DIFF
--- a/telemetry/vscode/src/parser.ts
+++ b/telemetry/vscode/src/parser.ts
@@ -42,7 +42,7 @@ export interface MetricDefinitionRoot {
 
 export function validateInput(fileText: string, fileName: string): MetricDefinitionRoot {
     try {
-        const schemaInput = readFileSync(path.join(__dirname, '../lib/telemetrySchema.json'), 'utf8')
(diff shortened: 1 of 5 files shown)


telemetry/vscode/src/parser.ts
```diff
@@ -42,7 +42,7 @@ export interface MetricDefinitionRoot {
 
 export function validateInput(fileText: string, fileName: string): MetricDefinitionRoot {
     try {
-        const schemaInput = readFileSync(path.join(__dirname, '../lib/telemetrySchema.json'), 'utf8')
+        const schemaInput = readFileSync(path.join(__dirname, '../../telemetrySchema.json'), 'utf8')
```
justinmk3, line 45: Don't require "package" step to run tests.

JadenSimon (approved):

PR #334, opened 2022-06-16
fix(parser): parser fails in consumers

justinmk3:
Problem:
39cb2fb900d2 changed this parser path to avoid needing to "package" the
project just to test it. But that path is based on assumptions about the
distributed artifact layout.

Solution:
Revert the path change and modify the "test" task so that it arranges
the "lib/" directory as expected by parser.ts.

## License

### DIFF
--- a/telemetry/vscode/src/parser.ts
+++ b/telemetry/vscode/src/parser.ts
@@ -42,7 +42,7 @@ export interface MetricDefinitionRoot {
 
 export function validateInput(fileText: string, fileName: string): MetricDefinitionRoot {
     try {
-        const schemaInput = readFileSync(path.join(__dirname, '../../telemetrySchema.json'), 'utf8')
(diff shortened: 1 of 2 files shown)


telemetry/vscode/src/parser.ts
```diff
@@ -42,7 +42,7 @@ export interface MetricDefinitionRoot {
 
 export function validateInput(fileText: string, fileName: string): MetricDefinitionRoot {
     try {
-        const schemaInput = readFileSync(path.join(__dirname, '../../telemetrySchema.json'), 'utf8')
+        const schemaInput = readFileSync(path.join(__dirname, '../lib/telemetrySchema.json'), 'utf8')
```
justinmk3, line 45: I should have known better ><

JadenSimon (approved):

PR #335, opened 2022-06-17
Update cloudwatch logs filter metric to indicate filter types separately

awschristou:
## Description

This change is a follow-up to #325 , to adjust the "filter logs" metric so that filter types are not mutually exclusive.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -46,11 +46,6 @@
             "description": "The duration of the operation in milliseconds"
         },
         {
-            "name": "filterBy",
-            "allowedValues": ["prefix", "time"],
-            "description": "Type of filter applied"
-        },
-        {
             "name": "result",
             "allowedValues": ["Succeeded", "Failed", "Cancelled"],
             "description": "The result of the operation"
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -60,6 +55,16 @@
             "allowedValues": ["editor", "viewer", "filesystem"],
             "description": "The IDE or OS component used for the action. (Examples: S3 download to filesystem, S3 upload from editor, ...)"
         },
+        {
+            "name": "isTextFilter",
+            "type": "boolean",
+            "description": "A text based filter was used"
+        },
+        {
+            "name": "isTimeFilter",
```
JadenSimon, line 64:
```suggestion
            "name": "hasTimeFilter",
```

Would be interesting to see what the common ranges are for filtering. Just don't know what the best way is to encode this.

JadenSimon (approved):

telemetry/definitions/commonDefinitions.json
shruti0085: Yeah I agree that'll be intereseting. Maybe having some default intervals enums as optional metadata field like [15 min, 30 min, 1 hr, 12 hr, 1day, more than a day]

shruti0085 (approved):

PR #352, opened 2022-08-10
[CodeWhisperer] add new telemetry definition aws_modifySetting

Will-ShaoHua:
Add new cwspr telemetry definition for tracking users enabling/disabling auto suggestion

## Checklist

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -355,6 +355,16 @@
             "description": "The id of the experiment being activated or deactivated"
         },
         {
+            "name": "settingId",
+            "type": "string",
+            "description": "The id of the setting being changed"
+        },
+        {
+            "name": "settingState",
+            "type": "string",
+            "description": "The state of the setting being changed to"
+        },
+        {
             "name": "toolId",
             "type": "string",
             "description": "The tool being installed",
@@ -374,12 +384,6 @@
             "name": "codewhispererAcceptedTokens",
             "type": "int",
             "description": "The metrics accepted on suggested CodeWhisperer code"
-        },
-        {
-            "name": "codewhispererAutomatedTriggerState",
-            "type": "string",
-            "description": "State of cwspr auto-suggestion",
-            "allowedValues": ["activated", "deactivated"]
         },
         {
             "name": "codewhispererAutomatedTriggerType",
@@ -1575,6 +1579,19 @@
             "passive": true
         },
         {
+            "name": "aws_modifySetting",
+            "description": "An setting was changed by users in the toolkit",
+            "metadata": [
+                {
+                    "type": "settingId"
+                },
+                {
+                    "type": "settingState",
+                    "required": false
+                }
+            ]
+        },
+        {
             "name": "deeplink_open",
             "description": "User requested that a resource be opened in the browser using the deeplink service",
             "metadata": [
@@ -1590,15 +1607,6 @@
                 }
             ],
             "passive": true
-        },
-        {
-            "name": "codewhisperer_autoSuggestionActivation",
-            "description": "User activate or deactivate automated trigger for CodeWhisperer code suggestion",
-            "metadata":  [
-                {
-                    "type": "codewhispererAutomatedTriggerState"
-                }
-            ]
         },
         {
             "name": "codewhisperer_codePercentage",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -376,6 +376,12 @@
             "description": "The metrics accepted on suggested CodeWhisperer code"
         },
         {
+            "name": "codewhispererAutomatedTriggerState",
+            "type": "string",
+            "description": "State of cwspr auto-suggestion",
+            "allowedValues": ["activated", "deactivated"]
+        },
+        {
             "name": "codewhispererAutomatedTriggerType",
             "type": "string",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1585,6 +1591,15 @@
             ],
             "passive": true
         },
+        {
+            "name": "codewhisperer_autoSuggestionActivation",
+            "description": "State of cwspr auto-suggestion",
```
andrewyuq, line 1596: Think something like "User activated or deactivated automated trigger for CodeWhisperer" is better

telemetry/definitions/commonDefinitions.json
```diff
@@ -375,6 +375,12 @@
             "type": "int",
             "description": "The metrics accepted on suggested CodeWhisperer code"
         },
+        {
+            "name": "codewhispererAutomatedTriggerState",
+            "type": "string",
+            "description": "State of cwspr auto-suggestion",
+            "allowedValues": ["activated", "deactivated"]
```
justinmk3, line 382: It's noisy to create a new type for every field in each metric. Is there an existing type that can be re-used? I see `experimentState` but that is named too specifically (which is similar to the problem we want to avoid here).

There is `enabled` https://github.com/aws/aws-toolkit-common/blob/2a44db2267931adfe47af6bb37bcbd98b223884c/telemetry/definitions/commonDefinitions.json#L208
but it is a boolean. Do you plan to add more values in the future?

JadenSimon: Another option is to add a generic metric like `aws_modifySetting` and add a single required field `settingId` and one optional field `settingState`. Both strings.

Will-ShaoHua: thanks Justin and Jaden for the great inputs, both of them work for me, will update it.

Will-ShaoHua: @JadenSimon, so the caller should look something similar to `AwsTelemetry.modifySetting("cwspr_autoSuggestionActivation", "activated")`, both of the fields will be pure strings, am I correct? Thanks!

JadenSimon: > with the option u mentioned, so the caller should look something similar to AwsTelemetry.modifySetting(settingId = "cwspr_autoSuggestionActivation", settingState = "activated"), both of the fields will be pure strings, am I correct? Thanks!

Yup. My suggestion is intended to be more re-usable while decreasing rigidity (which is both good and bad). I'd suggest using the same "namespace" for the setting id: `codewhisperer_autoSuggestionActivation` to keep things consistent.

Eventually I'd like to have a better way to express 'subtypes' for metrics. That way we can share high-level common definitions with increasing levels of detail per feature/component without losing out on the type-safety.

telemetry/definitions/commonDefinitions.json
```diff
@@ -354,6 +354,16 @@
             "type": "string",
             "description": "The id of the experiment being activated or deactivated"
         },
+        {
+            "name": "settingId",
+            "type": "string",
+            "description": "The id of the setting being changed"
```
JadenSimon, line 360: I think we should give an example in the description:
```suggestion
            "description": "The id of the setting being changed, e.g. codewhisperer_autoSuggestionActivation"
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -1568,6 +1578,19 @@
             ],
             "passive": true
         },
+        {
+            "name": "aws_modifySetting",
+            "description": "An setting was changed by users in the toolkit",
```
JadenSimon, line 1583:
```suggestion
            "description": "A setting was changed by users in the Toolkit. This metric can optionally provide the new state of the setting via settingState.",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -354,6 +354,16 @@
...
         },
+        {
+            "name": "settingId",
+            "type": "string",
+            "description": "The id of the setting being changed"
+        },
+        {
+            "name": "settingState",
+            "type": "string",
+            "description": "The state of the setting being changed to"
```
JadenSimon, line 365: Generic metrics are more likely to be used for a wide variety of purposes, some of which we probably don't want. While this is unlikely to be problem, we'll want to establish the intended use right away just in case.

```suggestion
            "description": "The state of the setting being changed to. This should not be recorded for free-form settings like file-system paths. Instead, stick to things like flags, numbers, and enums."
```

telemetry/definitions/commonDefinitions.json
Will-ShaoHua: noted, yea will update changes you suggested.
Thanks Jaden!

andrewyuq (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -1545,7 +1555,7 @@
         },
         {
             "name": "aws_experimentActivation",
-            "description": "An experiment was activated or deactivated in the toolkit",
+            "description": "An experiment was activated or deactivated in the Toolkit",
             "metadata": [
                 {
                     "type": "experimentId"
```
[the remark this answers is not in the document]
JadenSimon: Thinking we should keep it because it's specific to experiments while still being generic to features. Plus it's already being used, easier to let it be.

JadenSimon (approved):

PR #358, opened 2022-08-22
feat(codewhisperer): Add telemetry field of overall customer project size for security scan

fangwade007:
In order to collect data points about what's the P90 project size used for CWSPR security scan feature, so we can use it to scale our backend service, will need to add this additional field to emit the metrics.

## Types of changes

## Description
Add telemetry field of overall customer project size for security scan.

## Motivation and Context
Provide data point for backend scalability.

## Checklist

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -403,6 +403,11 @@
             "description": "How many lines of code being sent for security scan"
         },
         {
+            "name": "codewhispererCodeScanProjectBytes",
+            "type": "double",
+            "description": "The total size in bytes of customer project to perform security scan on"
+        },
+        {
             "name": "codewhispererCodeScanSrcPayloadBytes",
             "type": "int",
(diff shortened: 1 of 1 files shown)


zixlin7 (approved):

JadenSimon (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -402,6 +402,11 @@
             "type": "int",
             "description": "How many lines of code being sent for security scan"
         },
+        {
+            "name": "codewhispererCodeScanProjectBytes",
+            "type": "double",
```
justinmk3, line 407: why double instead of int? does "int" map to a smaller type?

is kilobytes a more appropriate unit (though ideally all of the "size" types would use the same unit)?

fangwade007: the main reason is that customer's project can be very large, unlike truncated payload size; and yes, we want to keep the same unit for size related fields

justinmk3 (approved):

PR #366, opened 2022-09-08
[codewhisperer]: Add telemetry for perceived latency

zixlin7:
## Types of changes

## Description
Add a new telemetry event to reflect user's perceived latency for seeing recommendations

## Checklist

## License


mr-lee (approved):

andrewyuq (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -1825,6 +1830,31 @@
                     "type": "codewhispererTriggerType"
                 }
             ]
+        },
+        {
+            "name": "codewhisperer_perceivedLatency",
```
JadenSimon, line 1835: Instead of creating a new definition, would it be possible to add `codewhispererPerceivedLatencyDuration` to some other event? `codewhisperer_userDecision` seems like a logical choice as it's only possible to pick a suggestion if you can see it. Or at least I would assume that's the case.

zixlin7: @JadenSimon `codewhisperer_userDecision` will be emitted for every recommendation(even the ones are not seen by users), but this metrics will only be emitted at most once for every `serviceInvocation` event, only if the first recommendation was shown to the user. So it's better for it to be separate.

JadenSimon: If this way is easier for y'all then it works for me. Just make sure it's emitted around the same time as `codewhisperer_userDecision` and marked passive.

telemetry/definitions/commonDefinitions.json
```diff
@@ -541,6 +541,11 @@
             "type": "string",
             "description": "The type of the user trigger to send request to CodeWhisperer service",
             "allowedValues": ["OnDemand", "AutoTrigger"]
+        },
+        {
+            "name": "codewhispererPerceivedLatencyDuration",
```
JadenSimon, line 546: Does it make sense to use the already defined `duration` attribute instead? We can put the description in the metric description.

JadenSimon (approved):

telemetry/definitions/commonDefinitions.json
zixlin7: updated, thanks!

JadenSimon (approved):

PR #367, opened 2022-09-15
[CodeWhisperer] add successCount in cwspr_codePercentage as serviceInvocation count

Will-ShaoHua:
Use existing type `successfulCount` to denote `cwspr_serviceInvocationCount` in metric `cwspr_percentage`

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the **README** document
- [ ] I have read the **CONTRIBUTING** document
- [ ] My code follows the code style of this project
- [ ] All new and existing tests passed

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1627,6 +1627,9 @@
                 {
                     "type": "reason",
                     "required": false
+                },
+                {
+                    "type": "successCount"
                 }
             ],
             "passive": true


telemetry/definitions/commonDefinitions.json
```diff
@@ -1627,6 +1627,9 @@
                 {
                     "type": "reason",
                     "required": false
+                },
+                {
+                    "type": "successCount"
```
leigaol, line 1632: How about successInvocationCount?

leigaol: Is this a required field?

Will-ShaoHua: sounds good, make more sense with successInvocationCOunt

Will-ShaoHua: yep, it's required, IMO

Will-ShaoHua: oh n,o there is no existing type `successInvocationCount`

Will-ShaoHua: https://github.com/aws/aws-toolkit-common/blob/master/telemetry/definitions/commonDefinitions.json#L374

Will-ShaoHua: I think `succssfulCount` should be the one we are looking for as per the description except we wanna create another type @leigaol

Will-ShaoHua: since we might want to use existing type as possible as we can

Will-ShaoHua: @JadenSimon @justinmk3

justinmk3: > use existing type as possible as we can

yes, thank you :)

JadenSimon: Yeah try to re-use things if possible, just don't force it. If something is different then let it be different.

We haven't really used `successCount` much but I think the original intent was to describe 'compressed' events e.g. instead of emitting 10 events for `s3_uploadObject`, emit 1 event with `successCount` set to 10. If `successCount` doesn't seem too ambiguous for your use-case then go ahead and use it. Otherwise I'd create a new field that is more specific.

JadenSimon (approved):

telemetry/definitions/commonDefinitions.json
Will-ShaoHua: @JadenSimon @justinmk3, noted, just think of a question, in this PR we added a `required` field `succssCount`, however the plugin versions in the marketplace doesn't have this field yet, will telemetry still consume these data without the required field ?

JadenSimon: You can make it required. 

If we were talking about the API for a service then making it required would be a breaking change. However, the telemetry backend is agnostic to these definitions and so backwards compatibility is only a concern for people looking at the data and not the service.

PR #372, opened 2022-09-22
Add exception handler for optional telemetry callback

shruti0085:
## Description

This change is a follow-up to PR #296 adding a guard against exceptions to the optional callback introduced to transform the metric properties prior to sending them to the server. 
Updates to telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/Generated/GeneratedCode.cs are autogenerated. 
## License

### DIFF
--- a/telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
+++ b/telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
@@ -56,7 +56,7 @@ namespace Test
                 datum.AddMetadata("awsAccount", payload.AwsAccount);
                 datum.AddMetadata("awsRegion", payload.AwsRegion);
 
-                datum = datum.InvokeTransform(telemetryLogger.Logger, transformDatum);
+                datum = datum.InvokeTransform(transformDatum);
 
                 metrics.Data.Add(datum);
                 telemetryLogger.Record(metrics);
@@ -100,7 +100,7 @@ namespace Test
                 datum.AddMetadata("awsAccount", payload.AwsAccount);
                 datum.AddMetadata("awsRegion", payload.AwsRegion);
 
-                datum = datum.InvokeTransform(telemetryLogger.Logger, transformDatum);
+                datum = datum.InvokeTransform(transformDatum);
 
                 metrics.Data.Add(datum);
                 telemetryLogger.Record(metrics);
@@ -154,7 +154,7 @@ namespace Test
 
                 datum.AddMetadata("runtime", payload.Runtime);
 
-                datum = datum.InvokeTransform(telemetryLogger.Logger, transformDatum);
+                datum = datum.InvokeTransform(transformDatum);
 
                 metrics.Data.Add(datum);
                 telemetryLogger.Record(metrics);
@@ -208,7 +208,7 @@ namespace Test
 
                 datum.AddMetadata("runtime", payload.Runtime);
 
-                datum = datum.InvokeTransform(telemetryLogger.Logger, transformDatum);
+                datum = datum.InvokeTransform(transformDatum);
 
                 metrics.Data.Add(datum);
                 telemetryLogger.Record(metrics);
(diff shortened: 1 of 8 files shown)


telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
```diff
@@ -56,9 +56,17 @@ namespace Test
                 datum.AddMetadata("awsAccount", payload.AwsAccount);
                 datum.AddMetadata("awsRegion", payload.AwsRegion);
 
-                if ((transformDatum != null))
+                try
```
awschristou, line 59: Another option would be to move this code to a new helper class in telemetry/csharp/AwsToolkit.Telemetry.Events/Core and make a static method that accepts a MetricDatum list and a transform, and returns a MetricDatum list, and then you would only need to generate the call to this method.

I'm okay with this current approach too.

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
```diff
@@ -27,6 +27,8 @@ public class DefinitionsBuilder
         private readonly CodeMethodReferenceExpression _debugAssert = 
             new CodeMethodReferenceExpression(new CodeTypeReferenceExpression(typeof(System.Diagnostics.Debug)), "Assert");
 
+        private readonly CodeTypeReferenceExpression _debuggerAttached = new CodeTypeReferenceExpression(typeof(System.Diagnostics.Debugger));
```
awschristou, line 30: This relates to "Debugger" not the Attached property, consider removing Attached from the name.

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
```diff
@@ -430,9 +432,35 @@ private CodeMemberMethod CreateRecordMetricMethodByDataClass(Metric metric)
...
+
+            catchClauseTransform.Statements.Add(new CodeExpressionStatement(
+                new CodeMethodInvokeExpression(
+                    new CodeFieldReferenceExpression(new CodeArgumentReferenceExpression("telemetryLogger"), "Logger"),
+                    "Error",
+                    new CodePrimitiveExpression("Error invoking transform function"),
+                    new CodeArgumentReferenceExpression("e"))
+            ));
+
+            // System.Diagnostics.Debug.Assert(!Debugger.IsAttached, "Error invoking transform function");
```
awschristou, line 449: Update the comment to "Debugger.IsAttached == false"

awschristou (approved):

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator.Tests/test-data/expectedCode-supplemental.txt
shruti0085: Updated - I like this idea.

telemetry/csharp/AwsToolkit.Telemetry.Events/Core/MetricDatumExtensionMethods.cs
```diff
@@ -57,5 +61,27 @@ public static void AddMetadata(this MetricDatum metricDatum, string key, int val
         {
             metricDatum.AddMetadata(key, value.ToString(System.Globalization.CultureInfo.InvariantCulture));
         }
+
+        /// <summary>
+        /// If the transform function isn't null, invoke it and assign metric datum to it's result
+        /// </summary>
+        public static MetricDatum InvokeTransform(this MetricDatum metricDatum, ILog logger,
```
awschristou, line 68: We don't need to pass in the logger. This class can just use its own.
Its okay if we don't test that something was logged. The most interesting test to have is that an exception does not escape form calling `InvokeTransform` (which would break the call stack and possibly crash the IDE)

awschristou (approved):

PR #394, opened 2022-12-01
Add CodeCatalyst definitions

JadenSimon:
## Types of changes

## Checklist

- [ ] I have read the **README** document
- [ ] I have read the **CONTRIBUTING** document
- [ ] My code follows the code style of this project
- [ ] All new and existing tests passed

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -583,7 +583,7 @@
         {
             "name": "userId",
             "type": "string",
-            "description": "Opaque AWS ID identifier"
+            "description": "Opaque AWS Builder ID identifier"
         },
         {
             "name": "codecatalyst_devEnvironmentWorkflowError",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -215,7 +215,7 @@
             "name": "credentialSourceId",
             "type": "string",
             "description": "Where credentials are stored or retrieved from",
-            "allowedValues": ["sharedCredentials", "sdkStore", "ec2", "ecs", "envVars", "other"]
+            "allowedValues": ["sharedCredentials", "sdkStore", "ec2", "ecs", "envVars", "awsId", "other"]
         },
         {
             "name": "credentialType",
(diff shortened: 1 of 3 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -215,7 +215,7 @@
             "name": "credentialSourceId",
             "type": "string",
             "description": "Where credentials are stored or retrieved from",
-            "allowedValues": ["sharedCredentials", "sdkStore", "ec2", "ecs", "envVars", "other"]
+            "allowedValues": ["sharedCredentials", "sdkStore", "ec2", "ecs", "envVars", "awsId", "other"]
```
JadenSimon, line 218: Based off https://github.com/aws/aws-toolkit-visual-studio-staging/pull/1231

JadenSimon:
Need to fix VS codegen. It can't handle underscores in the field names ?

justinmk3 (approved):

rli (approved):

awschristou (approved):
We've typically avoided using underscores in the type definitions (eg: `cloudWatchLogsPresentation`), but these values are in use now so its worth committing a version that represents the current state of the Toolkits.

It would be worth revisiting this to remove the underscores if possible. We could get code generation collisions if later on we define different types named like `foo_updateBar` and `foo_update_bar`

JadenSimon:
> It would be worth revisiting this to remove the underscores if possible. We could get code generation collisions if later on we define different types named like foo_updateBar and foo_update_bar

I've been leaning towards using the underscore to delineate namespaces moving forward because we already do it for the metric definitions. The types and definitions are both top-level identifiers and so they should follow the same pattern.  

Migrating telemetry data is not fun so this is something we should figure out soon. I'll write up a doc and then we can go from there.

telemetry/definitions/commonDefinitions.json
```diff
@@ -559,6 +560,40 @@
...
+            "description": "Locality of the Amazon CodeCatalyst update dev environment request (i.e., from the thin client or the local IDE instance)",
+            "allowedValues": [
+                "remote",
+                "local"
+            ]
+        },
+        {
+            "name": "userId",
+            "type": "string",
+            "description": "Opaque AWS ID identifier"
```
shruti0085, line 586: AWS Builder ID?

shruti0085 (approved):

PR #406, opened 2023-01-05
Define metrics for viewing and editing DynamoDB objects

awschristou:
I also swapped the position of dynamodb_fetchRecords and dynamodb_openTable so that the dynamodb metrics are ordered alphabetically.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -339,7 +339,7 @@
         {
             "name": "dynamoDbTarget",
             "allowedValues": ["table", "tableProperties", "tableStream"],
-            "description": "The type of entity referenced by a metric or operation"
+            "description": "The type of DynamoDB entity referenced by a metric or operation"
         },
         {
             "name": "resourceType",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -336,6 +336,11 @@
             "allowedValues": ["primary", "localSecondary", "globalSecondary"],
             "description": "The type of index being hit for the query/scan operation"
         },
+        {
+            "name": "dynamoDbTarget",
```
JadenSimon, line 340: CWL uses `cloudWatchResourceType` for a similar concept but in this case we seem to be combining a resource type with abstractions associated with it? I don't think we need to separate the two ideas necessarily as the idea of a resource is abstract anyway. But at least for the DBB service spec a table is clearly a resource while the properties of it are not. I'm just thinking here, no changes needed.

telemetry/definitions/commonDefinitions.json
```diff
@@ -336,6 +336,11 @@
...
+            "allowedValues": ["table", "tableProperties", "tableStream"],
+            "description": "The type of entity referenced by a metric or operation"
```
shruti0085, line 342: To make it a little clearer, can there be a callout for `dynamodb` based entity in the description ?

shruti0085 (approved):

shruti0085 (approved):

JadenSimon (approved):

PR #408, opened 2023-01-06
C# Telemetry Generator: add the reason field to all telemetry events

awschristou:
This change updates the C# Telemetry generator so that every generated event model contains an optional reason field, and that the metrics will emit this field if it has been set.

## License

### DIFF
--- a/telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
+++ b/telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
@@ -21,6 +21,13 @@ namespace Amazon.AwsToolkit.Telemetry.Events.Generator
         private const string AddMetadataMethodName = "AddMetadata";
         private const string InvokeTransformMethodName = "InvokeTransform";
 
+        // contains metadata fields that should be skipped when generating code.
+        // These fields are covered by the class BaseTelemetryEvent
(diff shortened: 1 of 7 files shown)


JadenSimon (approved):

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
```diff
@@ -21,6 +21,13 @@ public class DefinitionsBuilder
         private const string AddMetadataMethodName = "AddMetadata";
         private const string InvokeTransformMethodName = "InvokeTransform";
 
+        // contains metadata fields that should be skipped when generating code.
+        // These fields are covered by the class BaseTelemetryEvent
+        private static readonly string[] ImplicitFields = 
+        {
+            "reason",
+        };
```
JadenSimon, line 29: I guess reflection isn't too ergonomic in C#?

awschristou: I opted for the simplest/fastest route at this time. We could annotate the BaseTelemetryEvent class and then use reflection to determine these fields, but for now we have just the one.

JadenSimon: Oh you have to annotate it? Yeah that's not ergonomic.

PR #412, opened 2023-01-09
Define metrics for EC2 operations

awschristou:
This change defines EC2 related metrics for operations in the AWS Toolkit for Visual Studio

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1068,6 +1068,50 @@
             ]
         },
         {
+            "name": "ec2_createAmi",
+            "description": "Create an image from an EC2 Instance",
+            "metadata": [
+                { "type": "result" }
+            ]
+        },
+        {
+            "name": "ec2_editInstanceElasticIp",
+            "description": "Associate or disassociate an Elastic IP with an EC2 Instance",
+            "metadata": [
+                { "type": "result" }, 
+                { "type": "enabled", "required": false}
+            ]
+        },
+        {
+            "name": "ec2_editInstanceShutdownBehavior",
+            "description": "Adjust the shutdown behavior of an EC2 Instance",
+            "metadata": [
+                { "type": "result" }
+            ]
+        },
+        {
+            "name": "ec2_editInstanceTerminationProtection",
+            "description": "Adjust the termination protection of an EC2 Instance",
+            "metadata": [
+                { "type": "result" }, 
+                { "type": "enabled", "required": false}
+            ]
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1067,6 +1067,50 @@
...
+            "metadata": [
+                { "type": "result" }
+            ]
+        },
+        {
+            "name": "ec2_editInstanceElasticIp",
+            "description": "Associate or disassociate an Elastic IP with an EC2 Instance",
+            "metadata": [
+                { "type": "result" }, 
+                { "type": "enabled", "required": false}
```
JadenSimon, line 1082: So was this `enabled` field one of the reasons why we didn't want generic definitions? I could see how this might clash with something generic.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1067,6 +1067,50 @@
...
+            ]
+        },
+        {
+            "name": "ec2_editInstanceShutdownBehavior",
+            "description": "Adjust the shutdown behavior of an EC2 Instance",
+            "metadata": [
+                { "type": "result" }
```
JadenSimon, line 1089: We'll probably move `result` to be a common field soon. There's little reason why we _wouldn't_ want it considering that every event could potentially fail.

JadenSimon (approved):
I think these definitions would work fine following a similar pattern in https://github.com/aws/aws-toolkit-common/pull/406 after accounting for the "enablement" field. But it's not something I'm too worried about.

telemetry/definitions/commonDefinitions.json
awschristou: Later on we could realize that we want additional properties on specific events. By not going the generic metric approach, we avoid having a metric containing a higher list of (optional) metadata fields.

telemetry/definitions/commonDefinitions.json
awschristou: I was thinking the same thing today. I'm tempted to update the .NET generator to do this (much like #408), but would like to get these metrics hooked up to the Toolkit first.

telemetry/definitions/commonDefinitions.json
JadenSimon: > Later on we could realize that we want additional properties on specific events

This can still be done with generic events due to how we define metadata but it's not exactly intuitive for developers. It doesn't require doing anything special, it's just a different way of thinking about things.

PR #429, opened 2023-02-01
Define metric when a connection requires a browser login

shruti0085:
## Description
Defines a metric for when a connection requires a login using the browser

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -789,6 +789,14 @@
             "passive": true
         },
         {
+            "name": "aws_loginWithBrowser",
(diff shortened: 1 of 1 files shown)


JadenSimon (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -788,6 +788,14 @@
             ],
             "passive": true
         },
+        {
+            "name": "aws_loginWithBrowser",
```
justinmk3, line 792: there is also `aws_openUrl` . 

did we consider adding this as metadat on `aws_validateCredentials` instead? or introduce a `aws_login` which could have a flag that indicates the kind of login?

PR #433, opened 2023-02-09
feat(lsp): Port the states language service into the language server

jpinkney-aws:
This PR integrates the [states language service](https://github.com/aws/amazon-states-language-service) into the language server.

A few notes about this integration:
- The code is directly copied from [90e2eb6ddd05107df4b4bc2b04578ca5dcf1eb62](https://github.com/aws/amazon-states-language-service/tree/90e2eb6ddd05107df4b4bc2b04578ca5dcf1eb62)
- The only thing that has been directly changed is stepfunctions/snippets/error_handling and stepfunctions/snippets/states. They now export an array instead of json
- Minor changes have been made to types (removing some anys, etc) to fix build errors
- The code has been re-formatted according to our formatter rules
- Any eslint fixes to the code have been applied
- It currently only has support for `*.asl.json` and `*.asl.yml/*.asl.yaml` files

A couple observations/potential things to do:
- Introduce a factory-like thing around the configuration of language servers. We don't need to provide most the options every single time because they are almost always going to be the same (validation, hover, etc), we are really just changing schemas
- It may make sense to have one json/yaml language server per file type for maximum isolation
- Support detecting the contents of the file and "recognizing" the states language

There's an argument to be made about where this code should technically live. Each with their own pros and cons.

Everything living in this repo:
Pros:
- Centralization, one language server is a lot easier to find then many 
- Teams don't have to care about linting rules, setting up a new project, getting a repo, etc, we can give service teams an interface to implement like the [LanguageService](https://github.com/aws/aws-toolkit-common/blob/test-typescript-lsp/lsp/src/server/service.ts#L21). They implement it and then they have a fully functional language server for their service. In theory, we would also automagically hook up things like telemetry for them as well
- Simplification and de-duplication. Each language service implementation that the vscode toolkit uses seems to be doing the exact same things multiple times (setting up the language server, setting up the settings, etc, etc), this provides that in one spot
- Easier dependency management. Both the states language service and the ssm language service are using versions of the yaml language server, language server dependencies that are 2+ years old. In their current form those libraries are pretty hard to update and both the states language service and the ssm language service would have to do that upgrading independently. If there was a central place for defining where those dependencies are it would be easier to keep everything updated. This could also make it easier to trouble shoot where a single dependency is going wrong in the future

Cons:
- In this case, we are porting over functionality that already exists outside of this repo and is semi-maintained by another team (any they already publish an npm package) so we would have to ask them to contribute changes to this repository (not too big of a deal in exchange for that, their maintenance burden would be lower. They don't have to handle releases, they don't have to care about updating dependency versions, etc)
- Teams don't have any flexibility, they might need to work with us to define any new standards they want to implement, getting things merged and released faster, etc

## Testing
- All 147 tests provided by the states language service are passing at the service level (tests are instrumented through `yaml.ts` and `json.ts` in this PR)
- When running the VSCode client contained in this repo you get auto completion, validation, hover **without** the VSCode-YAML extension installed and the AWS-Toolkit-VSCode extension installed
 

## License


awschristou:
Language server code has moved to another repo, and has been removed #598. This PR can serve as a reference, but will be closed.

PR #436, opened 2023-02-09
doc: language service architecture doc

nkomonen-amazon:
This doc is implemented in mermaid.

Use https://mermaid.live for an improved editing experience.

See rendered diagram: https://github.com/nkomonen-amazon/aws-toolkit-common/blob/test-typescript-lsp/lsp/ARCHITECTURE.md

Signed-off-by: Nikolas Komonen <nkomonen@amazon.com>

## License

### DIFF
--- a/lsp/ARCHITECTURE.md
+++ b/lsp/ARCHITECTURE.md
@@ -1,0 +1,171 @@
+# Architechture
+
+## Language Service
+
+The following is a class diagram for the underlying Language Service.
+These are the components that enable users to provide a text document
+and get language support (hover, completion, ...) through Typescript code (as a Node module).
+
+This is separate from interfacing with it through a Language Server.
+
+### Example pseudocode:
+
+A quick example of how you would go from a text document
+to getting completion results.
+
+```typescript
+// Text document we want to do completion for
+incomingTextDocument: TextDocument = ...
+
+// Create language service strategy instances.
+// Note each strategy consumes the document parser, the strategy should know which parser to use.
+languageServiceStrategyA: LanguageServiceStrategy = new LanguageServiceStrategyA()
+languageServiceStrategyB: LanguageServiceStrategy = new LanguageServiceStrategyB()
+
+// Add strategies to the context
+languageServiceContext: LanguageServiceContext = ...
+languageServiceContenxt.addStrategy(languageServiceStrategyA)
+languageServiceContenxt.addStrategy(languageServiceStrategyB)
+
+// Resolve a specific Language Service instance from the context
+languageService: LanguageService = languageServiceContext.getLanguageService(incomingTextDocument)
+
+// Use the Language Service instance to do a completion
+completionResults = languageService.completion(incomingTextDocument, ...)
+```
+
+```mermaid
+classDiagram
+
+    
+
+    class JSONDocument {
+
+    }
+
+    note for JSONDocument "Implemented by node module:\nvscode-json-languageservice"
+
+    class YAMLDocument {
+
+    }
+
+    
+    JSONDocument <|.. YAMLDocument : Extends
+
+    class DocumentParserVisitor {
+        <<interface>>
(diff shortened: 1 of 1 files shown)


jpinkney-aws:
Are you able to attach a png to this PR that shows the diagram?

nkomonen-amazon:
> Are you able to attach a png to this PR that shows the diagram?

@jpinkney-aws I've linked to the rendered page in the description. This will be the latest diagram in the case it needs to be updated.

lsp/ARCHITECTURE.md
```diff
@@ -0,0 +1,110 @@
+# Architechture
+
+## Language Service
+
+The following is a class diagram for the underlying Language Service.
+These are the components that allow us to provide a text document and get
```
awschristou: Can you be more specific around what entity is "us" - does this mean language server to language service, Toolkit/IDE to language server, ...

lsp/ARCHITECTURE.md
```diff
@@ -0,0 +1,110 @@
...
+
+    class BuildspecService {
+
+    }
+
+    BuildspecService ..|> LanguageService : Implements
+
+    %% ---
+
+    %% This is provided by the VSCode node module
```
awschristou: While looking at the document, I was wondering which portions belong to what modules. It looks like mermaid doesn't support the concept of packages (which could contain entities like classes). What do you think about adding things like `note for TextDocument "Provided by package vscode-foo"` so that we see the comments in-diagram?

Another idea would be to style/color the classes along their different ownerships, but that could get cumbersome to manage.

Are any other interfaces/classes the domain of other packages, and could those be annotated as well?

lsp/ARCHITECTURE.md
nkomonen-amazon: Updated to add more clarity

lsp/ARCHITECTURE.md
```diff
@@ -0,0 +1,168 @@
...
+
+The following is a class diagram for the underlying Language Service.
+These are the components that enable users to provide a text document
+and get language support (hover, completion, ...) through Typescript code (as a Node module).
+
+This is separate from interfacing with it through a Language Server.
+
+### Example pseudocode:
+
+This is how
```
awschristou, line 13: The sentence cut off

lsp/ARCHITECTURE.md
```diff
@@ -0,0 +1,168 @@
...
+languageService: LanguageService = languageServiceContext.getLanguageService(incomingTextDocument)
+
+// Use the Language Service instance to do a completion
+completionResult = languageService.completion(incomingTextDocument, ...)
+```
+
+```mermaid
+classDiagram
+
+    class DocumentAST{
```
awschristou, line 42: Are these provided by an external node module or did we define and implement these? (if external, can we add a note?)

awschristou (commented):
This is looking good to me. @jpinkney-aws does this design accommodate for the use cases you had in mind for the language system?

jpinkney-aws:
Yeah, this is exactly what I had in mind. Any new file type just has to implement 1 interface for identifying the file, hook everything up and then from there the language server figures out where to route everything

lsp/ARCHITECTURE.md
awschristou: Heads up - this line appears to be truncated.

lsp/ARCHITECTURE.md
awschristou: We discussed this offline, the json and yaml types might become freestanding in the diagram (removing DocumentAst)

awschristou (approved):

lsp/ARCHITECTURE.md
nkomonen-amazon: Updated to add note explaining where the JSONDocument comes from (a separate node module)

PR #438, opened 2023-02-16
Set up packing command and GitHub CI

awschristou:
This change sets up a cheap CI to compile and test the language server code, as well as produce binaries that can be used for testing.

The change also makes some adjustments to the npm scripts, to add commands that will run the packager. The readme file was updated with things I discovered along the way.

## License


nkomonen-amazon (approved):
LGTM

PR #439, opened 2023-02-17
Define and update publish related metrics

shruti0085:
## Description
Defines and update metrics related to publish flows
* Adds `<service>_publishWizard` metric captured when user completes a service's publish wizard e.g. lambda deploy wizard. Captures where the action originated from(Except for ECS) and the duration of interaction with the wizard
* Updates deploy related metrics to capture duration of deploy

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -831,10 +831,23 @@
             "metadata": [
                 { "type": "result" },
                 { "type": "initialDeploy" },
+                { "type": "duration" },
                 { "type": "name", "required": false },
                 { "type": "framework", "required": false },
                 { "type": "xrayEnabled", "required": false },
-                { "type": "enhancedHealthEnabled", "required": false }
+                { "type": "enhancedHealthEnabled", "required": false },
+                { "type": "serviceType", "required": false },
+                { "type": "source" , "required": false}
+            ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -831,10 +831,23 @@
...
+                { "type": "serviceType", "required": false },
+                { "type": "source" , "required": false}
+            ]
+        },
+        {
+            "name": "beanstalk_publishWizard",
+            "description": "Called when user completes the Elastic Beanstalk publish wizard",
+            "metadata": [
+                { "type": "result" },
+                { "type": "duration" },
```
JadenSimon, line 848: Maybe add `duration` as a base field to everything similar to `reason`?

JadenSimon (approved):

telemetry/definitions/commonDefinitions.json
shruti0085: Do we see this getting potentially used for majority of our metrics in future? wondering for one-click actions like Copy/View?

awschristou: It would be reasonable for duration to be an optional base field. Such work would take place in a separate change though. (duration and result could be added to the base class in one pass)

shruti0085: Sounds good. I'll merge this and add those in a separate PR

PR #448, opened 2023-02-28
Add language server CI buildspec; package into bin folder

awschristou:
This change defines a buildspec file to be used as the Language Server CI. I also updated the package script to produce the binaries in a `bin` subfolder, so that the files are easier to identify as produced artifacts.

## License


JadenSimon (approved):

buildspec/languageServerTests.yml
```diff
@@ -0,0 +1,21 @@
+version: 0.2
```
justinmk3, line 1: Other files use "lsp" in the name, consider naming this file with the same pattern (`lspTests.yml`).

.github/workflows/lsp-ci.yaml
```diff
@@ -48,5 +48,5 @@ jobs:
             - name: Attach binaries
               uses: actions/upload-artifact@v3
               with:
-                name: langauge-server
-                path: lsp/awsdocuments-language-server-*
+                  name: langauge-server
+                  path: lsp/bin/awsdocuments-language-server-*
```
justinmk3, line 52: I downloaded the artifacts and I see:

```
awsdocuments-language-server-linux
awsdocuments-language-server-macos
awsdocuments-language-server-win.exe
```

That is very verbose. The files could be named `aws-lsp-xx` instead without any ambiguity.

buildspec/languageServerTests.yml
awschristou: lsp does not correctly describe the system, which is a language server that happens to handle the lsp protocol.

.github/workflows/lsp-ci.yaml
awschristou: This is the default behavior of the packager system. We will likely revisit the file names when we get closer to making final distribution bundles.

awschristou: For example, and TBD: we may have to differentiate x64 vs ARM binaries.

buildspec/languageServerTests.yml
justinmk3: LSP is the common moniker. Is there a risk of confusion? Is there a reasonable chance that any non-LSP lang server will be implemented?

PR #453, opened 2023-03-01
C# Telemetry Generator: add the duration field to all telemetry events

shruti0085:
## Description
This change updates the C# Telemetry generator so that every generated event model contains an optional duration field, and that the metrics will emit this field if it has been set.

Follows same pattern as PR #408
## License

### DIFF
--- a/telemetry/csharp/AwsToolkit.Telemetry.Events/Core/BaseTelemetryEvent.cs
+++ b/telemetry/csharp/AwsToolkit.Telemetry.Events/Core/BaseTelemetryEvent.cs
@@ -23,6 +23,12 @@ namespace Amazon.AwsToolkit.Telemetry.Events.Core
         /// </summary>
         public string Reason;
 
+        /// <summary>
+        /// Optional - The duration for the workflow associated with the metric 
+        /// This is often used in multi-step workflows to provide additional details about how long did the action take
+        /// </summary>
+        public double Duration;
+
         public DateTime? CreatedOn;
         public double? Value;
         public string AwsAccount;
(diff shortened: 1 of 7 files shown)


telemetry/csharp/AwsToolkit.Telemetry.Events/Core/BaseTelemetryEvent.cs
```diff
@@ -23,6 +23,12 @@ public abstract class BaseTelemetryEvent
         /// </summary>
         public string Reason;
 
+        /// <summary>
+        /// Optional - The duration for the workflow associated with the metric 
+        /// This is often used in multi-step workflows to provide additional details about how long did the action take
+        /// </summary>
+        public double Duration;
```
awschristou, line 30: Discussed offline: we want to handle this field like an optional (similar to the `Value` property). This means it isn't emitted into json payloads if there isn't a value specified.

awschristou (approved):

justinmk3:
+5000 lines for some extra fields and extra metrics doesn't seem sustainable. Looks like the generated code could be reduced a lot.

awschristou:
> +5000 lines for some extra fields and extra metrics doesn't seem sustainable. Looks like the generated code could be reduced a lot.

This change had about 20 relevant lines of code changed. The rest goes to supporting tests and illustrating the generated code. What is your concern with the generated code? It does not have to be elegant or minimal, and has served us well for three years now.

justinmk3:
How much code is generated from the existing telemetry definitions? I can't tell from just this PR, but 5k lines of code for a small change, raises the question of how much is the total generated code. If it's 10k that's not a problem, but depending on how it scales as we gain more metrics, 1M could slow down builds (in CI and locally). Git, antivirus, compilers eventually do run into limits.

awschristou:
The resulting NuGet package is currently 40 kb (https://www.nuget.org/packages/AWS.Toolkit.Telemetry.Events). Unless you're planning on tripling the number of metrics defined week over week, I think we're okay for a while.

The 5k diff is somewhat misleading. The generated code file could probably be added to .gitignore if the PR delta size is a concern. I find it is an easy way to visually inspect what the generator output is looking like.

PR #457, opened 2023-03-10
Add language service overview to Architecture

awschristou:
## License


lsp/ARCHITECTURE.md
```diff
@@ -1,5 +1,19 @@
 # Architechture
```
jpinkney-aws, line 1: nit: 
```suggestion
# Architecture
```

jpinkney-aws (commented):
I'm almost wondering if it would be easier to start calling language services -> intellisense services or something like that. I know a lot of people have been tripped up by the service/server words

lsp/ARCHITECTURE.md
awschristou: Thank you!

JadenSimon (approved):

JadenSimon:
> I'm almost wondering if it would be easier to start calling language services -> intellisense services or something like that. I know a lot of people have been tripped up by the service/server words

I'm somewhat indifferent on the naming as it seems like there will always be a lot of overlap.

lsp/ARCHITECTURE.md
```diff
@@ -1,4 +1,18 @@
-# Architechture
+# Architecture
+
+The application consists of two components: the language server, and the language services.
+
+The **Language Server** is the containing application. It hosts language services, implements the Language server protocol (LSP), and queries information from language services in order to resepond to LSP requests.
```
justinmk3, line 5: The server responds to queries. The client requests from the server. 

This definition makes it sound like the server is the client.

lsp/ARCHITECTURE.md
```diff
@@ -1,4 +1,18 @@
...
+
+The **Language Services** are document-specific implementations that understand how to parse and interpret certain file types. These are used by the language server in order to provide LSP clients with details that power language assistance functionality.
+
+## Language Server
+
+![Langauge server architecture diagram](docs/images/language-server-architecture.svg)
```
justinmk3, line 11: The diagram makes more sense. It seems like "service" is a concept we introduced? Calling it a "provider" would be less confusing. Then we would define them as:

```
The **Language Server** is implements the Language server protocol (LSP), and queries information from document-specific providers in order to resepond to LSP requests.

The **Language Providers** implement analysis of various document types. These are used by the language server to serve LSP client requests.
```

jpinkney-aws: Service is a concept we introduced so it's not mentioned anywhere in the LSP. It's because the [json](https://github.com/microsoft/vscode-json-languageservice) intellisense implementation is in a different repository than the actual server implementation so we started calling the json intellisense provider the language service and the thing that takes in the LSP requests and redirects them to the service as the language server. The YAML language server ran with that idea as well, except for it has the server + service implementation in the same repo. Other language servers don't have this kind of idea though. It's unique to JSON/YAML iirc

That's why in my previous comment I wondered if we should pick a different name, since quite a few stakeholders were getting confused between service and server

justinmk3: > Service is a concept we introduced

Yup, just mentioning that little bit of info removes a lot of confusion here.

PR #459, opened 2023-03-13
feat(codewhisperer): add new telemetry for import recommendation

leigaol:
## Problem

Add new telemetry for feature: import recommendation

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -546,6 +546,21 @@
             "description": "Number of references the particular suggestion is referenced with."
         },
         {
+            "name": "codewhispererSuggestionImports",
+            "type": "string",
+            "description": "The list of import statement for a particular suggestion"
+        },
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -545,6 +545,21 @@
             "type": "int",
             "description": "Number of references the particular suggestion is referenced with."
         },
+        {
+            "name": "codewhispererSuggestionImports",
```
justinmk3, line 549: Why add more and more fields to `codewhisperer_userDecision`? Would it make sense to introduce a new metric, e.g. `codewhisperer_userDecisionImport` ?

That would avoid needing many of these special-case fields, and it makes the metrics more meaningful. For example, instead of one `codewhisperer_userDecision` metric with 20 fields, there could be `count` field and similar common fields.

justinmk3 (approved):

PR #478, opened 2023-05-03
telemetry: allow filter-related fields on cloudwatchlogs_open

justinmk3:
`cloudwatchlogs_open` generally covers all cases where cloudwatch logs are fetched in various ways, which may include filter parameters. Adding these fields to `cloudwatchlogs_open` allows that metric to be used generically instead of the special-case `cloudwatchlogs_filter`.

After this change, `cloudwatchlogs_filter` metric is redundant/unnecessary.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1125,7 +1125,7 @@
         },
         {
             "name": "cloudwatchlogs_filter",
-            "description": "Filters a CloudWatch Logs entity. Deprecated: use cloudwatchlogs_open instead.",
+            "description": "Filters a CloudWatch Logs entity. (Essentially a subset of cloudwatchlogs_open)",
             "metadata": [
                 { "type": "result" },
                 { "type": "cloudWatchResourceType" },
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1052,6 +1052,8 @@
                 { "type": "cloudWatchResourceType" },
                 { "type": "cloudWatchLogsPresentation", "required": false },
                 { "type": "serviceType", "required": false },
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1123,7 +1125,7 @@
         },
         {
             "name": "cloudwatchlogs_filter",
-            "description": "Filters a CloudWatch Logs entity.",
+            "description": "Filters a CloudWatch Logs entity. Deprecated: use cloudwatchlogs_open instead.",
```
JadenSimon, line 1128: I believe `cloudwatchlogs_filter` is emitted in VS when users filter logs already opened in the viewer. So this event is still relevant because that information isn't captured in `cloudwatchlogs_open`. 

```suggestion
            "description": "Emitted when applying a filter to an already opened CloudWatch Logs entity",
```

shruti0085: +1, VS provides the ability to filter logs once the stream/group has been opened for viewing in the IDE.

justinmk3: > `cloudwatchlogs_filter` is emitted in VS when users filter logs already opened in the viewer

We also have that feature in vscode, but https://github.com/aws/aws-toolkit-vscode/pull/3410 changes the logic to capture that as `cloudwatchlogs_open` with `source=Editor` + these new fields.

JadenSimon: > with source=Editor + these new fields.

Oh I didn't notice that we're changing `source` to reflect this. That works too although it may not be as obvious.

> The API calls are essentially the same, no? The only difference is the "source" of the action.

I wouldn't focus too much on the API calls. The events should reflect user actions/experiences. I'd rather have duplication than ambiguous data.

JadenSimon (approved):

PR #480, opened 2023-05-03
telemetry: add generic `requestId`

JadenSimon:
**This field is intentionally "lossy" in order to accommodate a wide variety of scenarios**

I know I said that unambiguous data is favorable over duplication [here](https://github.com/aws/aws-toolkit-common/pull/478#discussion_r1184051015). That's still true. But data with a small amount of uncertainty is still more useful than no data. As much as I would like to differentiate every single API call made during a workflow, that's just not practical right now.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -90,7 +90,7 @@
         {
             "name": "requestServiceType",
             "type": "string",
-            "description": "The service identifier associated with a request. Values should have the same shape as `serviceType`. This is distinct from `serviceType` which describes the originator of the request rather than the request itself."
+            "description": "Per-request service identifier. Unlike `serviceType` (which describes the originator of the request), this describes the request itself."
         },
         {
             "name": "requestId",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -88,6 +88,11 @@
             "description": "The name of the AWS service acted on. These values come from the AWS SDK. To find them in the JAVA SDK search for SERVICE_NAME in each service client, or look for serviceId in metadata in the service2.json"
         },
         {
+            "name": "requestId",
+            "type": "string",
+            "description": "A generic request ID field. The semantics are contextual based off of other fields (e.g. `serviceType`). For example, an event with `serviceType: s3` means that the request ID is associated with an S3 API call. Events that cover mutliple API calls should use the request ID of the most recent call."
+        },
+        {
             "name": "name",
             "type": "string",
             "description": "A generic name metadata"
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -87,6 +87,16 @@
             "type": "string",
             "description": "The name of the AWS service acted on. These values come from the AWS SDK. To find them in the JAVA SDK search for SERVICE_NAME in each service client, or look for serviceId in metadata in the service2.json"
         },
+        {
+            "name": "requestServiceId",
+            "type": "string",
+            "description": "The service identifier associated with a request. Values should have the same shape as `serviceType`. This is distinct from `serviceType` which describes the originator of the request rather than the reqeust itself."
+        },
+        {
+            "name": "requestId",
```
JadenSimon, line 96: We could join the request ID + service ID into `${serviceId}:${requestId}` to be more defensive but now we're assuming a specific shape whereas keeping them separate allows another field to act as a disambiguator.

telemetry/definitions/commonDefinitions.json
```diff
@@ -87,6 +87,16 @@
             "type": "string",
             "description": "The name of the AWS service acted on. These values come from the AWS SDK. To find them in the JAVA SDK search for SERVICE_NAME in each service client, or look for serviceId in metadata in the service2.json"
         },
+        {
+            "name": "requestServiceId",
+            "type": "string",
+            "description": "The service identifier associated with a request. Values should have the same shape as `serviceType`. This is distinct from `serviceType` which describes the originator of the request rather than the request itself."
```
justinmk3, line 93: if this is compatible with `serviceType`, would it help intuition to name it `requestServiceType` ?

telemetry/definitions/commonDefinitions.json
```diff
@@ -87,6 +87,16 @@
             "type": "string",
             "description": "The name of the AWS service acted on. These values come from the AWS SDK. To find them in the JAVA SDK search for SERVICE_NAME in each service client, or look for serviceId in metadata in the service2.json"
         },
+        {
+            "name": "requestServiceType",
+            "type": "string",
+            "description": "The service identifier associated with a request. Values should have the same shape as `serviceType`. This is distinct from `serviceType` which describes the originator of the request rather than the request itself."
```
justinmk3, line 93:
```suggestion
            "description": "Per-request service identifier. Unlike `serviceType` (which describes the originator of the request), this describes the request itself."
```

telemetry/definitions/commonDefinitions.json
JadenSimon: Yeah might as well. I had it that way at first but it felt weird to call it a 'type'.

justinmk3 (approved):

PR #483, opened 2023-05-16
LSP: Fix bug in matching strategy when working with multiple providers

awschristou:
When we have multiple language providers, where is a bug where the selector does not choose the first applicable provider. This change fixes that bug.

I also found that I couldn't place breakpoints on service files, so I updated launch.json to fix that problem.

## License

### DIFF
--- a/lsp/src/service/registry/registry.ts
+++ b/lsp/src/service/registry/registry.ts
@@ -26,6 +26,13 @@ export class LanguageServiceRegistry implements LanguageServiceContext {
     }
 
     private async getMatchingStrategy(textDocument: TextDocument): Promise<LanguageServiceStrategy | undefined> {
-        return Promise.any(this.strategies.filter(strategy => strategy.isMatch(textDocument)))
+        for (const strategy of this.strategies) {
+            const isMatch = await strategy.isMatch(textDocument)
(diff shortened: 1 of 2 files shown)


lsp/src/service/registry/registry.ts
```diff
@@ -26,6 +26,13 @@ export class LanguageServiceRegistry implements LanguageServiceContext {
     }
 
     private async getMatchingStrategy(textDocument: TextDocument): Promise<LanguageServiceStrategy | undefined> {
-        return Promise.any(this.strategies.filter(strategy => strategy.isMatch(textDocument)))
+        for (const strategy of this.strategies) {
+            const isMatch = await strategy.isMatch(textDocument)
```
justinmk3, line 30: What is the difference vs `strategies.filter` ? Maybe just `Promise.any` isn't needed?

nkomonen-amazon (approved):

lsp/src/service/registry/registry.ts
justinmk3: > The updated code also exits after the first successful match. We don't need to run all of the remaining evaluations.

sure, can use `some()` instead of filter()

PR #490, opened 2023-06-01
C# Telemetry Generator: add the locale field to all telemetry events

rbbarad:
This change updates the C# Telemetry generator so that every generated event model contains an optional locale field, and that the metrics will emit this field if it has been set.

Follows same pattern as PR https://github.com/aws/aws-toolkit-common/pull/408

Manually modified files:
- BaseTelemetryEvent.cs
- DefintionsBuilder.cs
- GeneratedSupplementalCodeTests.cs

GeneratedCode.cs and SupplementalCode.cs files were generated by running the GeneratedSupplementalCodeTests.cs

__Testing:__
Ran AwsToolkit.Telemetry.Events.Tests and AwsToolkit.Telemetry.Events.Generator.Tests
## License

### DIFF
--- a/telemetry/csharp/AwsToolkit.Telemetry.Events/Core/BaseTelemetryEvent.cs
+++ b/telemetry/csharp/AwsToolkit.Telemetry.Events/Core/BaseTelemetryEvent.cs
@@ -29,6 +29,11 @@ namespace Amazon.AwsToolkit.Telemetry.Events.Core
         /// </summary>
         public double? Duration;
 
+        /// <summary>
+        /// Optional - Language-related user preference information. Examples: en-US, en-GB, etc.
+        /// </summary>
+        public string Locale;
+
         public DateTime? CreatedOn;
         public double? Value;
         public string AwsAccount;
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -313,6 +313,11 @@
             "description": "Language-specific identification. Examples: v4.6.1, netcoreapp3.1, nodejs12.x. Not AWS Lambda specific. Allows for additional details when other fields are opaque, such as the Lambda runtime value 'provided'."
         },
         {
+            "name": "locale",
+            "type": "string",
+            "description": "Language-related user preference information. Examples: en-US, en-GB, etc."
+        },
+        {
             "name": "insightsDialogOpenSource",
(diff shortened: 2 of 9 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -312,6 +312,11 @@
             "type": "string",
             "description": "Language-specific identification. Examples: v4.6.1, netcoreapp3.1, nodejs12.x. Not AWS Lambda specific. Allows for additional details when other fields are opaque, such as the Lambda runtime value 'provided'."
         },
+        {
+            "name": "locale",
+            "type": "string",
+            "description": "Language-related user preference information. Examples: en-US, en-GB, etc."
```
floralph, line 318: Any pros/cons to using allowedValues here?

rbbarad: I avoided that since the allowedValues could go pretty large over time. Open to changing that if desired

awschristou: I agree with making this one open ended. That avoids a treadmill maintaining an enumeration

telemetry/definitions/commonDefinitions.json
```diff
@@ -2238,6 +2243,10 @@
                 {
                     "type": "reason",
                     "required": false
+                },
+                {
+                    "type": "locale",
```
awschristou, line 2248: What do you think about making this an optional field at the code generator level, that way all metrics have the option of emitting this?

floralph (approved):

telemetry/definitions/commonDefinitions.json
rbbarad: ah i like that idea - will make that adjustment

telemetry/csharp/AwsToolkit.Telemetry.Events/Core/BaseTelemetryEvent.cs
awschristou, line 1: It just occurred to me that we should update https://github.com/aws/aws-toolkit-common/blob/main/telemetry/telemetryformat.md#global-arguments with properties like this. Can you please add locale with your PR and we'll worry about the catching up the other fields separately?

FYI @JadenSimon

awschristou (approved):

awschristou (approved):

PR #497, opened 2023-06-07
Proof of Concept: Reusable JSON and Yaml servers powered by JSON Schemas

awschristou:

This change contains proof of concept implementations of JSON (`JsonSchemaServer`) and Yaml (`YamlSchemaServer`) language servers, that operate on a JSON Schema. The classes can be instantiated with a url to a JSON Schema, and it will provide completion/validation based on that schema. A follow-up change will illustrate how these are leveraged. These servers are not optimized (or de-duplicated) at this time.

To support the proof of concept, code has been added to the core library. The `AwsLanguageService` interface is a concept that will be explored further as we grow out the monorepo. This may provide a boundary that contains a document's (or language)'s domain knowledge, without coupling to the LSP Connection.

## License

### DIFF
--- a/lsp/core/aws-lsp-core/src/language-service/awsLanguageService.ts
+++ b/lsp/core/aws-lsp-core/src/language-service/awsLanguageService.ts
@@ -1,8 +1,12 @@
 import { Position, Range, TextDocument, TextEdit } from 'vscode-languageserver-textdocument'
 import { CompletionList, Diagnostic, FormattingOptions, Hover } from 'vscode-languageserver-types'
 
-// Where possible, this interface should follow the (VS Code) JSON Language Server's interface, which is close
-// to the core (VS Code) language server implementations.
+// Where possible, this interface should follow the (VS Code) JSON Language Server's interface,
+// (https://github.com/microsoft/vscode-json-languageservice/blob/v5.3.5/src/jsonLanguageService.ts#L38)
+// which closely resembles the core (VS Code) language server implementations.
+// We would prefer to leverage truly core types (like those from vscode-languageserver-textdocument),
+// since the JSON Language Service is a specific product and not a generalized interface type,
+// we have an explicit interface here for use in language services (and consumers).
 export interface AwsLanguageService {
     isSupported(document: TextDocument): boolean
     doValidation(textDocument: TextDocument): PromiseLike<Diagnostic[]>
--- a/lsp/core/aws-lsp-core/src/language-service/awsLanguageService.ts
+++ b/lsp/core/aws-lsp-core/src/language-service/awsLanguageService.ts
@@ -1,0 +1,12 @@
+import { Position, Range, TextDocument, TextEdit } from 'vscode-languageserver-textdocument'
+import { CompletionList, Diagnostic, FormattingOptions, Hover } from 'vscode-languageserver-types'
+
+// Where possible, this interface should follow the (VS Code) JSON Language Server's interface, which is close
+// to the core (VS Code) language server implementations.
+export interface AwsLanguageService {
+    isSupported(document: TextDocument): boolean
+    doValidation(textDocument: TextDocument): PromiseLike<Diagnostic[]>
+    doComplete(textDocument: TextDocument, position: Position): PromiseLike<CompletionList | null>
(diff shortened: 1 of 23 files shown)


lsp/core/aws-lsp-core/src/language-service/awsLanguageService.ts
```diff
@@ -0,0 +1,12 @@
+import { Position, Range, TextDocument, TextEdit } from 'vscode-languageserver-textdocument'
+import { CompletionList, Diagnostic, FormattingOptions, Hover } from 'vscode-languageserver-types'
+
+// Where possible, this interface should follow the (VS Code) JSON Language Server's interface, which is close
+// to the core (VS Code) language server implementations.
+export interface AwsLanguageService {
```
justinmk3, line 6: is this a subset of the vscode interface? would `Pick<>` be useful here?

lsp/core/aws-lsp-json-common/src/language-server/jsonLanguageService.ts
```diff
@@ -0,0 +1,60 @@
...
+export type JsonLanguageServiceProps = {
+    defaultSchemaUri?: string
+    schemaProvider?: SchemaProvider
+}
+
+export class JsonLanguageService implements AwsLanguageService {
+    private jsonService: LanguageService
+
+    constructor(private readonly props: JsonLanguageServiceProps) {
+        this.jsonService = getLanguageService({
```
justinmk3, line 15: So this (and the yamlservice) are just wrappers around 3P sevices. That is a key insight that would make the web of classes easier to digest. Could perhaps just be mentioned in the docstring on each class, to start with.

lsp/core/aws-lsp-json-common/src/language-server/jsonSchemaServer.ts
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Options, provided by language server to `onInitialize` callback.
+     */
+    // protected options?: InitializeParams;
+    /**
+     * Text documents manager (see https://github.com/microsoft/vscode-languageserver-node/blob/main/server/src/common/textDocuments.ts#L68)
+     */
+    protected documents = new TextDocuments(TextDocument)
+
+    protected jsonService: AwsLanguageService
```
justinmk3, line 28: and the "server" wraps the "service"...

justinmk3 (approved):
These incremental PRs are super helpful for following along, thank you!

Naming nite: "common" in `aws-lsp-yaml-common` seems unnecessary, it could just be `aws-lsp-yaml`. Same for the json variant.

lsp/core/aws-lsp-core/src/language-service/awsLanguageService.ts
awschristou: It is a subset of the [JSON Language Service](https://github.com/microsoft/vscode-json-languageservice/blob/main/src/jsonLanguageService.ts#L38)

I feel like the interface I linked to is the top level "json language server" interface. If it was a more generalized interface in the language server package (instead of the json service package) I'd be inclined to use that as the central type. I think we'll want to explore this interface concept further during the experiment phase though.

lsp/core/aws-lsp-json-common/src/language-server/jsonLanguageService.ts
awschristou: Thanks, this is a great call out. I will add the comments

lsp/core/aws-lsp-core/src/language-service/awsLanguageService.ts
justinmk3: Makes sense. Might be worth linking to that in the docstring, that kind of historical context can help comprehension.

awschristou:
> Naming nit: "common" in `aws-lsp-yaml-common` seems unnecessary, it could just be `aws-lsp-yaml`. Same for the json variant.

We'll definitely revisit this. As we do our exploration, these supporting packages will be pretty fluid, and we'll get a better sense of what packages we need (and need to publish). Once we have a better sense of the packages we can refine the names (prior to publishing).

PR #498, opened 2023-06-07
Buildspec servers

awschristou:

This change demonstrates how we can create a language server around a CodeBuild Buildspec JSON Schema, building on the proof of concept reusable components defined in #497 .

Three separate language servers have been created:
- one that applies buildspec logic to JSON files (BuildspecJsonServer)
- one that applies buildspec logic to Yaml files (BuildspecYamlServer)
- one that applies buildspec logic to both JSON and Yaml files (BuildspecServer)

BuildspecJsonServer and BuildspecYamlServer shows how we can leverage something that just knows how to wrap a JSON Schema. We don't know yet if we'd want to host disparate language servers like these in the IDE, which is where BuildspecServer comes in. We don't necessarily want to compose language servers together, because there will be collisions at the lsp connection callback level. BuildspecServer illustrates how one level lower (language services) may be a good reusability layer.

To support this change, MutuallyExclusiveLanguageService (naming is hard) shows how we can compose AwsLanguageService implementations in order to produce something that knows how to handle JSON and YAML files.

In a follow up change, I will restore the sample language clients, which will allow us to run and debug this language server while developing in the repo.

## License

### DIFF
--- a/lsp/core/aws-lsp-core/src/index.ts
+++ b/lsp/core/aws-lsp-core/src/index.ts
@@ -1,5 +1,6 @@
 // todo : IDE-10874 : move core support files into this package from pre-monorepo locations
 export * from './content/schemaProvider'
 export * from './language-service/awsLanguageService'
+export * from './language-service/mutuallyExclusiveLanguageService'
 export * as completionItemUtils from './util/completionItemUtils'
 export * as textDocumentUtils from './util/textDocumentUtils'
--- a/lsp/server/aws-lsp-buildspec/src/language-server/buildSpecServer.ts
+++ b/lsp/server/aws-lsp-buildspec/src/language-server/buildSpecServer.ts
@@ -1,0 +1,134 @@
+import {
+    AwsLanguageService,
+    MutuallyExclusiveLanguageService,
+    SchemaProvider,
+    completionItemUtils,
+    textDocumentUtils,
+} from '@lsp-placeholder/aws-lsp-core'
+import { JsonLanguageService } from '@lsp-placeholder/aws-lsp-json-common'
+import { YamlLanguageService } from '@lsp-placeholder/aws-lsp-yaml-common'
+import {
+    Connection,
+    InitializeParams,
+    InitializeResult,
+    TextDocumentSyncKind,
+    TextDocuments,
+} from 'vscode-languageserver'
+import { TextDocument } from 'vscode-languageserver-textdocument'
+
+export type BuildspecServerProps = {
+    connection: Connection
+    defaultSchemaUri: string
+    schemaProvider: SchemaProvider
+}
+
+/**
+ * This is a demonstration language server that handles both JSON and YAML files according to the
+ * CodeBuild BuildSpec JSON Schema.
+ *
+ * This illustrates how we can wrap LSP Connection calls around a composition of different language services.
+ * In this case, the composition consists of a JSON processor and a YAML processor.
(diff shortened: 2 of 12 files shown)


lsp/core/aws-lsp-core/src/index.ts
```diff
@@ -1,5 +1,6 @@
 // todo : IDE-10874 : move core support files into this package from pre-monorepo locations
 export * from './content/schemaProvider'
```
nkomonen-amazon, line 2: nit: we should stick with camelCase folder names to keep things consistent

lsp/server/aws-lsp-buildspec/src/language-server/buildSpecServer.ts
```diff
@@ -0,0 +1,134 @@
...
+    protected documents = new TextDocuments(TextDocument)
+
+    protected buildSpecService: AwsLanguageService
+
+    protected connection: Connection
+
+    constructor(private readonly props: BuildspecServerProps) {
+        this.connection = props.connection
+
+        this.buildSpecService = new MutuallyExclusiveLanguageService([
```
nkomonen-amazon, line 44: We should be creating the Buildspec Service outside of the server as its own class. We have to keep in mind it is likely the service will be used outside the context of a Language Server (the same way we use the yaml language service ourselves).

Just a `BuildspecService` class which extends MutuallyExclusiveLanguageService, so users don't have to think of how to build an equivalent Buildspec service instance

nkomonen-amazon (approved):

lsp/core/aws-lsp-core/src/index.ts
awschristou: I've taken a note to revisit this as we wrap up explorations and start formalizing things in the lsp repo

lsp/server/aws-lsp-buildspec/src/language-server/buildSpecServer.ts
awschristou: I agree with this - it fits our need to mix and match services so that we aren't fix to a one server to one service approach. I'm planning on making a PR that adds a CloudFormation server in the style you mentioned, for the purpose of comparing these design concepts.

lsp/server/aws-lsp-buildspec/src/language-server/buildSpecServer.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/**
+ * This is a demonstration language server that handles both JSON and YAML files according to the
+ * CodeBuild BuildSpec JSON Schema.
+ *
+ * This illustrates how we can wrap LSP Connection calls around a composition of different language services.
+ * In this case, the composition consists of a JSON processor and a YAML processor.
+ */
+export class BuildspecServer {
```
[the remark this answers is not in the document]
awschristou: I've taken a note to revisit this. My thought is that language servers will be separate implementations, and we'll want those to be created with full access to the connection. A base server is a convenience, but we should be mindful of additional complexity.

lsp/server/aws-lsp-buildspec/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+    "compilerOptions": {
+        "module": "UMD",
```
JadenSimon, line 3: Why UMD? Can we not use native ES modules?

PR #500, opened 2023-06-08
Buildspec binary; Relocate sample language clients

awschristou:

This change demonstrates how we can create a standalone language server binary that can be integrated into IDEs. A binary is created using the Buildspec server created in https://github.com/aws/aws-toolkit-common/pull/498.

This change also relocates the sample language client extensions to their new location in the monorepo. The sample VS Code client has been adjusted to run using its updated pathing, and a launch profile has been set up that is capable of running the buildspec server.

## License


lsp/tsconfig.json
```diff
@@ -12,6 +12,12 @@
     },
     "files": [],
     "references": [
+        {
+            "path": "./app/aws-lsp-buildspec-binary"
```
justinmk3, line 16: I see we have separate binaries for buildspec and other json/yaml-based DSLs. Did we not plan to have a single binary that can serve different kinds of json/yaml-based DSLs? At least until we run into limitations thereof.

awschristou: We have an investigation spike planned to assess the impacts of running multiple language servers, so for now I've created two separate binaries (one for each of buildspec and cloudformation, which will be in a follow-up PR) which will provide the means for experimenting

justinmk3 (approved):

PR #502, opened 2023-06-08
CloudFormation language server and binary

awschristou:

This change provides a second demonstration language server and binary that can be integrated into IDEs. This language server is based on the CloudFormation JSON Schema.

This server's implementation differs from the Buildspec implementation (from https://github.com/aws/aws-toolkit-common/pull/500), for the purpose of comparison. In this instance, the service is injected into the server. This looks like it gives us more design flexibility in reusing and composing the services to suit the needs of where they will be hosted/integrated.

## License

### DIFF
--- a/lsp/.vscode/launch.json
+++ b/lsp/.vscode/launch.json
@@ -14,7 +14,7 @@
             }
         },
         {
-            "name": "Debug BuildSpec Server in VSCode",
+            "name": "BuildSpec Server",
             "type": "extensionHost",
             "request": "launch",
             "runtimeExecutable": "${execPath}",
@@ -26,7 +26,7 @@
             "preLaunchTask": "npm: compile"
         },
         {
-            "name": "Debug CloudFormation Server in VSCode",
+            "name": "CloudFormation Server",
             "type": "extensionHost",
             "request": "launch",
             "runtimeExecutable": "${execPath}",
--- a/lsp/.vscode/launch.json
+++ b/lsp/.vscode/launch.json
@@ -26,6 +26,18 @@
             "preLaunchTask": "npm: compile"
         },
         {
+            "name": "Debug CloudFormation Server in VSCode",
+            "type": "extensionHost",
+            "request": "launch",
+            "runtimeExecutable": "${execPath}",
+            "args": ["--extensionDevelopmentPath=${workspaceFolder}/client/vscode"],
+            "outFiles": ["${workspaceFolder}/client/vscode/out/**/*.js"],
+            "env": {
+                "LSP_SERVER": "${workspaceFolder}/app/aws-lsp-cloudformation-binary/out/index.js"
+            },
+            "preLaunchTask": "npm: compile"
+        },
+        {
             "type": "node",
             "request": "launch",
             "preLaunchTask": "watch",
--- a/lsp/app/aws-lsp-cloudformation-binary/package.json
+++ b/lsp/app/aws-lsp-cloudformation-binary/package.json
@@ -1,0 +1,19 @@
+{
+    "name": "@lsp-placeholder/aws-lsp-cloudformation-binary",
+    "version": "0.0.1",
+    "description": "CodeBuild CloudFormation Language Server Binary",
+    "main": "out/index.js",
+    "bin": {
+        "awsdocuments-language-server": "./out/index.js"
+    },
+    "scripts": {
+        "compile": "tsc --build",
+        "package-x64": "pkg --targets node18-linux-x64,node18-win-x64,node18-macos-x64 --out-path bin --compress GZip ."
+    },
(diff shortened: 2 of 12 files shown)


lsp/.vscode/launch.json
```diff
@@ -25,6 +25,18 @@
             },
             "preLaunchTask": "npm: compile"
         },
+        {
+            "name": "Debug CloudFormation Server in VSCode",
```
justinmk3, line 29: is "in VSCode" is redundant in a vscode launch config name? "Debug" also is not normally mentioned

lsp/app/aws-lsp-cloudformation-binary/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "name": "@lsp-placeholder/aws-lsp-cloudformation-binary",
+    "version": "0.0.1",
+    "description": "CodeBuild CloudFormation Language Server Binary",
+    "main": "out/index.js",
+    "bin": {
+        "awsdocuments-language-server": "./out/index.js"
+    },
+    "scripts": {
+        "compile": "tsc --build",
+        "package-x64": "pkg --targets node18-linux-x64,node18-win-x64,node18-macos-x64 --out-path bin --compress GZip ."
```
justinmk3, line 11: why isn't this just a task in the main `aws-lsp-cloudformation` projec? why another project needed?

lsp/app/aws-lsp-cloudformation-binary/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+        "module": "UMD",
+        "target": "ES2021",
+        "moduleResolution": "node",
+        "lib": ["ES2021"],
+        "rootDir": "./src",
+        "outDir": "./out",
+        "declaration": true,
+        "sourceMap": true,
+        "skipLibCheck": true,
+        "forceConsistentCasingInFileNames": true,
```
justinmk3, line 12: the more projects we have, the more of this kind of configuration we have to paste and maintain in many places

lsp/app/aws-lsp-cloudformation-binary/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "@lsp-placeholder/aws-lsp-cloudformation-binary",
```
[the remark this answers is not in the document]
awschristou: The current thought is that we will have different language server artifacts for integration into browsers and into the Toolkits.

The browsers will need a library that exports classes, which are directly consumed by the front end developer (this is @lsp-placeholder/aws-lsp-cloudformation in this concept).

The Toolkits will need free-standing applications, which will be run on the host systems (this is @lsp-placeholder/aws-lsp-cloudformation-binary).

We still have some exploration work to do to accommodate the different constraints and requirements from each of these host points, which will in turn influence how we design for packages, and what would go into each package. Committing these changes into the monorepo is intended to give us a starting point for that exploration work (churn is expected).

Also: naming is hard :)

lsp/.vscode/launch.json
awschristou: We can sculpt the entry names over time. We could plausibly have launch tasks that open up the VS or JetBrains sample clients. I think that is not likely, but is why I was explicit with putting "... in VSCode" in the name.

lsp/app/aws-lsp-cloudformation-binary/package.json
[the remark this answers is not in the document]
justinmk3: But those are just different types of build targets, why do they require a separate project?

lsp/.vscode/launch.json
justinmk3: if we have jetbrains etc tasks, then "in Jetbrains" makes sense. but don't need to over-verbose the common case to support that

lsp/app/aws-lsp-cloudformation-binary/package.json
justinmk3: sure, those differences can be captured as a build task that produces a separate target. why is a separate project needed? this is a lot of fragmentation, which makes code sharing harder (already presenting in the form of copy-pasted tsconfig.json, etc)

lsp/.vscode/launch.json
awschristou: That's reasonable. I've updated the names

lsp/app/aws-lsp-cloudformation-binary/package.json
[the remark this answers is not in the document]
awschristou: This build target concept is something I'm not as familiar with in TS/JS. As we get more developers into this repo, we'll be able to refine this appropriately.

justinmk3 (approved):

PR #505, opened 2023-06-12
Add middleware for retrieving content, add http handler as first implementation

awschristou:

This change introduces a middleware pipeline (`ContentRequestMiddleware`, `SchemaProviderBuilder`) that allows language servers to request content from a location. This allows us to mix and match loading from different places like local filesystems, in memory caches, and over http, based on the needs/limitations of a language server host.

This change includes one handler implementation (`HttpHandler`), which loads content from http(s). More handlers will be added in upcoming changes, keeping this change smaller.

The middleware has been hooked into the CloudFormation server. Loading from http on each request is not practical or efficient -- additional handlers will be implemented and added to this middleware pipeline in upcoming PRs, which will allow servers to cache content.

The PR delta appears large, but most of it is in the package json lock file, since testing package dependencies have been added. `npm run test` has been added to the monorepo root, which will run all available tests from all monorepo packages.

## License


lsp/core/aws-lsp-core/src/content/contentRequestMiddleware.ts
```diff
@@ -0,0 +1,39 @@
+import { URI } from 'vscode-uri'
+
+/**
+ * Middleware pipeline handler for requesting content.
+ * Handlers are chained together so that a content requestor makes a single request call.
+ *
+ * Implementations have no knowledge of the calling pipeline, or of what other handlers are in the pipeline.
+ * An implementation may perform pre-processing on the request, and then call the delegate.
+ * An implementation may perform post-processing on the result returned by a delegate.
+ * If an implementation is capable of handling the request, it may return the result without calling the delegate.
```
floralph, line 10: Doesn't this break the first assertion in this paragraph?  This implies that the handlers can make some assumption of the other downstream handlers, that they're not doing meaningful work.

For example, if there is a handler downstream that does some cross-cutting/bookkeeping concern, it would get skipped if the handler didn't continue passing the request down the pipeline even if it assumes it handled it completely.

Is there a way for an eng to immediately know what handlers can perform this short-circuiting behavior when composing the pipeline (without having to read through every handler's code)?

lsp/core/aws-lsp-core/src/content/handlers/httpHandler.ts
```diff
@@ -0,0 +1,29 @@
...
+} from '../contentRequestMiddleware'
+
+export class HttpHandler implements ContentRequestMiddleware {
+    constructor(readonly httpContentDownloader: HttpRequester = new DefaultHttpRequester()) {}
+
+    /**
+     * Downloads the content of a HTTP uri.
+     *
+     * @param uri
+     * @param eTag
```
floralph, line 16: Should this be "next"?

lsp/core/aws-lsp-core/src/content/schemaProviderBuilder.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        const handler = sut.build()
+        const result = await handler(sampleUri)
+
+        expect(result).to.equal('foo')
+    })
+
+    it('delegates to next handler', async () => {
+        sut.addHandler(createDelegatingMiddleware())
+        sut.addHandler(createMiddlewareReturning('foo'))
```
floralph, line 46: Minor:  The tricky thing with this test is if the order were to be reversed by accident in SchemaProviderBuilder then the last handler returning 'foo' would always be called and returned first which will pass the test.  Is there a way for this test to ensure that the other handler was in fact called?

lsp/core/aws-lsp-core/src/content/handlers/httpHandler.ts
awschristou: Oops, thanks -- I'll update this

lsp/core/aws-lsp-core/src/content/schemaProviderBuilder.test.ts
awschristou: Yes, good catch.

lsp/core/aws-lsp-core/src/content/schemaProviderBuilder.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        sut.addHandler(createMiddlewareReturning('bar'))
+
+        const handler = sut.build()
+        const result = await handler(sampleUri)
+
+        expect(result).to.equal('foo')
+    })
+
+    function createMiddlewareReturning(contentResponse: string): ContentRequestMiddleware {
```
nkomonen-amazon, line 68: Should add another argument, `uri?: string` and add in logic to only return a response if the uri is matched. Will address the case ralph mentioned

lsp/core/aws-lsp-core/src/content/schemaProviderBuilder.ts
```diff
@@ -0,0 +1,34 @@
+import { URI } from 'vscode-uri'
+import { ContentRequestMiddleware, ContentRequestMiddlewareDelegate } from './contentRequestMiddleware'
+import { SchemaProvider } from './schemaProvider'
```
nkomonen-amazon, line 3: I think we should move away from the 'schema' naming scheme since this logic ultimately applies to URIs as a whole.

floralph (approved):

lsp/core/aws-lsp-core/src/content/schemaProviderBuilder.ts
awschristou: I think that is a good idea, we can encapsulate the "schema resolver" terminology that is used in the json/yaml servers. I will do a renaming pass as a separate PR.

lsp/core/aws-lsp-core/src/content/contentRequestMiddleware.ts
awschristou: The expectation is that one handler will ultimately produce the core response payload. Other handlers in the stack may handle a different aspect of the request relating to the core response payload.

lsp/core/aws-lsp-core/src/content/schemaProviderBuilder.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export class SchemaProviderBuilder {
+    private readonly handlers: ContentRequestMiddleware[] = []
+
+    public addHandler(handler: ContentRequestMiddleware): SchemaProviderBuilder {
+        this.handlers.push(handler)
+        return this
+    }
+
+    public build(): SchemaProvider {
```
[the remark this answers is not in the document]
awschristou: A pipeline gives us more flexibility. For example, we can have a handler that uses a cache if the content is considered up to date, otherwise delegates. Then we can have another handler that obtains (from http, or elsewhere). Both of these could be used for the same scheme. Rather than having one large implementation, we can separate these concerns, and even use them in different orchestrations. Peek ahead to #506 for an example of this.

PR #506, opened 2023-06-12
Add cached content and file loading handlers

awschristou:
# DRAFT
This change is a follow-up to #505 

https://github.com/aws/aws-toolkit-common/pull/505 introduced a middleware that can retrieve content for language servers, and provided an http(s) implementation. This change adds new middleware handlers, so that language servers can:
- load content from files (`FileHandler`)
- load content from http(s) and cache it, relying on etags to indicate when the cached content is outdated (handler: `CachedContentHandler`, cache: `UriCacheRepository`)

To support this change, pre-monorepo code has been reused and moved around:
- LanguageServerCacheDir became DefaultUriCacheLocation
- Time became TimeProvider
- resolve.ts (and its tests) has been removed. Its implementation has been split into the middleware handlers (and cache) now available

## License


lsp/core/aws-lsp-core/src/content/handlers/cachedContentHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+    async get(uri: URI, next: ContentRequestMiddlewareDelegate): Promise<ContentRequestResponse> {
+        if (!uri.scheme.startsWith('http')) {
+            // we're only caching from online sources
+            return next(uri)
+        }
+
+        const cachedMetadata = await this.cacheRepository.getContentMetadata(uri)
+        const currentETag = await this.getETag(uri)
+
+        if (this.isCacheStale(cachedMetadata) === false && cachedMetadata!.eTag === currentETag) {
```
floralph, line 53: Is something special happening with this bool rather than to use `!this.isCacheStale(cachedMetadata)`?

lsp/core/aws-lsp-core/src/content/handlers/cachedContentHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+    private async getETag(uri: URI): Promise<string | undefined> {
+        const response = await this.httpRequester.request(uri.toString(), {
+            type: 'HEAD',
+        })
+
+        return getETag(response.headers)
+    }
+
+    /** Checks if the cache has timed out using the last updated time. */
+    private isCacheStale(contentMetadata: ContentMetadata | undefined): boolean {
```
floralph, line 83: Why mix cache staleness handling into handler rather than cache classes?  Is there any cache eviction workflow that cleans up stale files (assuming they're over 30 minutes and not requested again)?

lsp/core/aws-lsp-core/src/content/handlers/fileHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+import {
+    ContentRequestMiddleware,
+    ContentRequestMiddlewareDelegate,
+    ContentRequestResponse,
+} from '../contentRequestMiddleware'
+const readFile = promisify(fs.readFile)
+
+export class FileHandler implements ContentRequestMiddleware {
+    async get(uri: URI, next: ContentRequestMiddlewareDelegate): Promise<ContentRequestResponse> {
+        if (!(uri.scheme == 'file')) {
```
floralph, line 13: For my own edification, as it's been awhile and I know JS has some duck typing funkiness, is there a difference between `!(uri.scheme == 'file')` and `uri.scheme != 'file'`?

lsp/core/aws-lsp-core/src/http/requesters.ts
```diff
@@ -24,5 +24,5 @@ export class DefaultHttpRequester implements HttpRequester {
 }
 
 export function getETag(headers: HttpRequestHeaders): string {
-    return headers.eTag as string
+    return (headers.eTag as string) ?? (headers.ETag as string) ?? (headers.etag as string)
```
floralph, line 27: Wow, would a case-insensitive search of Object.keys make sense here in case other variations such as Etag or ETAG come in?

lsp/core/aws-lsp-core/src/content/cache/uriCacheRepository.ts
```diff
@@ -0,0 +1,156 @@
...
+        await this.ensureCacheDirectoryExists()
+        await writeFile(absolutePath, text)
+    }
+
+    private async ensureCacheDirectoryExists(): Promise<void> {
+        await mkdir(this.cachedUrisDirPath, { recursive: true })
+    }
+
+    private hashUri(uri: URI): string {
+        return createHash('sha256').update(uri.toString()).digest('hex')
```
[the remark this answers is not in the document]
awschristou: Good question - this is used for producing a unique local cache path. There isn't a need for strong crypto (we aren't encoding secret data here), and the cache can be frequently accessed. I will change this to SHA-1, which appears to be 10-20% faster than sha256.

lsp/core/aws-lsp-core/src/content/cache/uriCacheRepository.ts
```diff
@@ -0,0 +1,156 @@
...
+        return this.getContentFileText(contentFileName)
+    }
+
+    /** Gets the content of a uri from the cache */
+    private async getContentFileText(contentFileName?: string): Promise<string | undefined> {
+        if (contentFileName === undefined) {
+            return
+        }
+
+        await this.ensureCacheDirectoryExists()
```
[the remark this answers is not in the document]
awschristou: I wasn't sure if exists would throw when any part of the path doesn't exist. I put a check to guard all file accesses.
I plan to leave this since these get accessed pretty frequently over the lifetime of a language server, so a directory (and its contents) will be pretty quickly re-populated.

lsp/core/aws-lsp-core/src/content/handlers/cachedContentHandler.ts
awschristou: You're correct, the intent is "this is not stale content (and the etag hasn't changed)"

lsp/core/aws-lsp-core/src/content/handlers/cachedContentHandler.ts
awschristou: My thinking was that the definition of stale should be defined by something that wants to use a cache, rather than by the cache.

There isn't anything today that would auto-delete files from the cache, but this could be added at a later date if needed.

lsp/core/aws-lsp-core/src/content/handlers/fileHandler.ts
awschristou: I don't know how that happened, I must have overlooked this during a refactor -- I have changed this to use `!==`

lsp/core/aws-lsp-core/src/http/requesters.ts
awschristou: Probably, we can optimize and improve this going forward as we use new urls that surface other casings.

lsp/core/aws-lsp-core/src/content/cache/uriCacheRepository.ts
```diff
@@ -0,0 +1,156 @@
...
+
+        const absolutePath = path.join(this.cachedUrisDirPath, contentFileName)
+        if (!(await exists(absolutePath))) {
```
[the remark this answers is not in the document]
floralph: Ok, this sounds reasonably safe then.  Checking for exist then acting on file is susceptible to a race condition.  A try/catch to react to the missing file (with or without an exists check) would avoid this.

floralph (approved):

lsp/core/aws-lsp-core/src/content/handlers/cachedContentHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Middleware that will use a cached version of content if the online version has not been updated.
+ * Newer content that has been downloaded will be cached.
+ *
+ * This avoids unnecessary downloads.
+ *
+ * Cached content is considered stale after {@link thirtyMinutesInMillis}
+ * and is refreshed.
+ */
+export class CachedContentHandler implements ContentRequestMiddleware {
```
justinmk3, line 26: Does this have logging or some other visibility into its decisions for troubleshooting

lsp/core/aws-lsp-core/src/content/cache/uriCacheRepository.ts
```diff
@@ -0,0 +1,159 @@
+import { createHash } from 'crypto'
+import * as fs from 'fs'
```
justinmk3, line 2: Will these 'fs' imports need to be revisited for supporting the browser? Or will there be a different impl of uriCacheRepository for browser?

lsp/core/aws-lsp-core/src/content/handlers/cachedContentHandler.ts
awschristou: TBD - We currently aren't using this component, and won't be for a while.

lsp/core/aws-lsp-core/src/content/cache/uriCacheRepository.ts
awschristou: TBD - these might get patched for browser use, or likely an alternate impl will be used. This sort of thing will likely become the responsibility of the "runtime" component, handled by a separate team. Our current focus is not on JSON Schemas, so we won't be working in this space for a while.

PR #510, opened 2023-06-13
Rename "Schema resolving" terminology to "URI resolving"

awschristou:
This change renames some classes and files to more generalized terms.

## License


lsp/core/aws-lsp-core/src/content/uriResolver.ts
```diff
@@ -2,6 +2,6 @@
  * The schema request service is used to fetch schemas. If successful, returns a resolved promise with the content of the schema.
  * In case of an error, returns a rejected promise with a displayable error string.
  */
-export interface SchemaProvider {
+export interface UriResolver {
```
justinmk3, line 5: does the docstring need an update

lsp/core/aws-lsp-core/src/content/uriResolverBuilder.ts
```diff
@@ -1,22 +1,22 @@
...
 
 /**
- * Creates a middleware pipeline for resolving requests for a schema.
+ * Creates a middleware pipeline for resolving requests for content from a uri.
  * Middleware handlers will be called in the order they are added, until a
  * handler returns a response. At this point, the handlers in the call stack
  * have a chance to handle/process the response.
  */
-export class SchemaProviderBuilder {
+export class UriResolverBuilder {
```
justinmk3, line 11: Why does a "resolver" need a "builder", can the indirection be eliminated?

justinmk3 (approved):
Definitely an improvement

lsp/core/aws-lsp-core/src/content/uriResolver.ts
awschristou: Yes, thanks for that catch

lsp/core/aws-lsp-core/src/content/uriResolverBuilder.ts
awschristou: The builder is a way of creating a middleware pipeline that chains uri resolvers together into a composite UriResolver. You can use UriResolver without the builder though. Open to suggestions for alternative names for UriResolverBuilder.

justinmk3: can the resolver be composable without being wrapped in a builder?

awschristou: Not very nicely. The code would look more complicated [in places where you would define the composition](https://github.com/aws/aws-toolkit-common/pull/510/files#diff-3bc3c866c395540756bac81fec325f2fa410073ce82467181fe68a65a723ad8fR24), and at the place where we [hand off the resolver into third party code](https://github.com/aws/aws-toolkit-common/pull/510/files#diff-5b6b83b0b1531d2a705ec63b9fd21cadcddad7c6b60375f62458a46073ccf6e2R20)

PR #520, opened 2023-06-19
Update Sample VS Client to integrate multiple LSPs

awschristou:
This change updates the sample VS LSP integration so that we can work with multiple language servers without so much boilerplate code. It also contains sample code to illustrate how to add support for JSON files, which are more strongly handled directly by Visual Studio.

## License


lsp/client/visualStudio/IdesLspPoc/ContentDefinitions/JsonContentType.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.VisualStudio.LanguageServer.Client;
+using Microsoft.VisualStudio.Utilities;
+using System.ComponentModel.Composition;
+
+namespace IdesLspPoc.ContentDefinitions
+{
+    // This is used to associate our language server with JSON files
+    // JSON handling is sort of reserved by VS, this is the only way I found that activates our LSP and uses it with .json files
```
floralph, line 8: Was MS forthcoming about this and were other content types listed or did you have to stumble upon this and search around a bit?  Curious how big an issue this might be for other content types.

lsp/client/visualStudio/IdesLspPoc/LspClient/BuildspecLspClient.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Name of Language Client; displayed to user
+        /// For example, if the LSP writes logs to an output window, this is where they will appear
+        /// </summary>
+        public override string Name => "AWS Toolkit language client for Buildspec Templates";
+
+        protected override string GetServerWorkingDir()
+        {
+            // to try using this extension, update dir to wherever your lsp service executable is
+            return @"C:\code\aws-toolkit-common\lsp\app\aws-lsp-buildspec-binary\bin";
```
floralph, line 31: Is this path always going to be hardcoded or will it be pulled from config somewhere?

lsp/client/visualStudio/IdesLspPoc/LspClient/CloudFormationLspClient.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    // this one "just works" to associate this LSP with yaml files
+    [ContentType("yaml")]
+
+    // for some reason, we have associate LSPs with JSON using a custom type - see JsonContentType for comments
+    [ContentType(JsonContentType.ContentTypeName)]
+
+    // [ContentType(BuildSpec.ContentType)] // Only add this if you're supporting custom file extensions
+    [Export(typeof(ILanguageClient))]
+    public class CloudFormationLspClient : ToolkitLspClient
```
floralph, line 20: Given the ContentType declarations here match that of the BuildspecLspClient above, how does VS know which to pick or does the VS extension code know and explicitly instantiate the correct one?  I'm guessing with the use of ExportAttribute, this may not be the latter case.

lsp/client/visualStudio/IdesLspPoc/ContentDefinitions/JsonContentType.cs
```diff
@@ -0,0 +1,28 @@
...
+    // See FAQ: https://learn.microsoft.com/en-us/visualstudio/extensibility/adding-an-lsp-extension?view=vs-2022#faq
+    public static class JsonContentType
+    {
+        // This also becomes the languageId provided to the language server
+        public const string ContentTypeName = "placeholder-test-json";
```
floralph, line 13: This will get a better/clearer name in the future?

floralph (approved):

lsp/client/visualStudio/IdesLspPoc/ContentDefinitions/JsonContentType.cs
awschristou: I had to stumble upon this through trial and error. I'm guessing each file type is implemented within Visual Studio as an internal extension, and there is likely some deviation for the behavior around each type.

lsp/client/visualStudio/IdesLspPoc/LspClient/BuildspecLspClient.cs
awschristou: This is currently written up in the lsp/client/visualStudio readme.md file. We'll want to clean this up to have something easy to use while iterating on the language servers though.

lsp/client/visualStudio/IdesLspPoc/LspClient/CloudFormationLspClient.cs
awschristou: Both classes are registering their respective language server to be associated with json and yaml files. Visual Studio handles multiple languages servers getting concurrently registered to a file type.

lsp/client/visualStudio/IdesLspPoc/ContentDefinitions/JsonContentType.cs
awschristou: Yes -- this is illustrative of how we will likely have to define some language ids if we want to codify logic using language ids in the language server.

PR #521, opened 2023-06-22
Proof of concept language server that lists S3 buckets

awschristou:
This change adds a concept language server that provides S3 bucket names as completion suggestions. It works on files that are named `*.s3.json`

This language server is being submitted so that we can perform additional experimentation like providing language servers with credentials, verifying browser compatibility, etc.

## License

### DIFF
--- a/lsp/tsconfig.json
+++ b/lsp/tsconfig.json
@@ -31,10 +31,16 @@
             "path": "./server/aws-lsp-cloudformation"
         },
         {
+            "path": "./server/aws-lsp-s3"
+        },
+        {
             "path": "./app/aws-lsp-buildspec-binary"
         },
         {
             "path": "./app/aws-lsp-cloudformation-binary"
+        },
+        {
+            "path": "./app/aws-lsp-s3-binary"
         }
     ]
 }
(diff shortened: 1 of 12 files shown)


lsp/tsconfig.json
```diff
@@ -30,11 +30,17 @@
...
+            "path": "./server/aws-lsp-s3"
+        },
         {
             "path": "./app/aws-lsp-buildspec-binary"
         },
         {
             "path": "./app/aws-lsp-cloudformation-binary"
+        },
+        {
+            "path": "./app/aws-lsp-s3-binary"
```
justinmk3, line 43: Looks like many projects. And the size of the package-lock.json in this PR is a hint about the dependency explosion we will need to manage with all of these micro-projects.

justinmk3 (approved):
A canonical skeleton/scaffold is very useful, nice.

lsp/tsconfig.json
awschristou: The cool thing about how an npm monorepo works is that there is a single package-lock.json (and single node_modules) folder that serves as the composite for all projects in the monorepo. Each project still uses what it needs (without bundling the other packages that it doesn't reference), but you don't have redundant node_modules folders for each project.

justinmk3: Nice!

PR #526, opened 2023-07-03
Proof of concept language server that provides CodeWhisperer recommendations

awschristou:
This change adds a concept language server that provides code recommendations from CodeWhisperer. The recommendations are returned through code completions and inline completion.

The sample VS Code extension has been updated to support this language server when editing typescript files. This is accomplished by introducing an inline completion provider, which wraps the concept language server.

At this time, inline completion is not a part of the language server protocol. VS Code has an extensibility API for inline completion, and a proposal is part of the next protocol version 3.18 (https://github.com/microsoft/language-server-protocol/pull/1673). This proposal is modelled after the extensibility API.

A likely implementation of the protocol is proposed in https://github.com/microsoft/vscode-languageserver-node/pull/1190, and portions of this change leverage the shapes defined in that proposal. This way, if or when inline completion is a part of the spec, we could update code to reference `vscode-languageserver` related types with little effort.

This concept uses IAM credentials with CodeWhisperer as a way to get up and iterating quickly. In future work we will want to switch over to using bearer tokens, like the AWS Toolkit for VS Code does.

The CodeWhisperer service client was produced using the AWS Toolkit for VS Code's client generator (https://github.com/aws/aws-toolkit-vscode/blob/master/scripts/build/generateServiceClient.ts)

This language server is being submitted so that we can perform additional experimentation such as:
- verifying browser compatibility
- exploring the UX around surfacing code recommendations
- future bearer token support

## License


lsp/client/vscode/src/futureTypes.ts
```diff
@@ -0,0 +1,45 @@
...
+    StaticRegistrationOptions,
+    TextDocumentPositionParams,
+    TextDocumentRegistrationOptions,
+    WorkDoneProgressOptions,
+    WorkDoneProgressParams,
+} from 'vscode-languageclient'
+
+/**
+ * Inline completion is not a part of the language server protocol.
+ * It is being proposed at this time.
```
nkomonen-amazon, line 13:
```suggestion
 * It is being proposed at this time (https://github.com/microsoft/language-server-protocol/pull/1673).
```

nkomonen-amazon (approved):

lsp/server/aws-lsp-codewhisperer/src/client/codewhispererclient.d.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * THIS FILE IS AUTOGENERATED BY 'generateServiceClient.ts'.
```
justinmk3, line 2: 👀

lsp/package.json
```diff
@@ -15,7 +15,7 @@
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
         "clean": "ts-node ./script/clean.ts",
-        "compile": "tsc --build --verbose",
+        "compile": "tsc --build --verbose && npm run postcompile --workspaces --if-present",
```
justinmk3, line 18: does npm implicitly run `postcompile` after `compile` https://docs.npmjs.com/cli/v9/using-npm/scripts#pre--post-scripts ? If so, it would seem redundant to call `postcompile` from `compile`.

lsp/server/aws-lsp-codewhisperer/src/client/codewhisperer.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ServiceConfigurationOptions } from 'aws-sdk/lib/service'
+const apiConfig = require('./service-2.json')
+import CodeWhispererClient = require('./codewhispererclient')
+
+export function createCodeWhispererClient(options: ServiceConfigurationOptions): CodeWhispererClient {
+    return createService(options) as CodeWhispererClient
+}
+
+function createService(options: ServiceConfigurationOptions): Service {
+    const client = new Service({ apiConfig, ...options } as any)
```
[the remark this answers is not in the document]
awschristou: I tried a variety of ways to instantiate this service client, and this was the only way I could get to work. I suspect this code may get some updates in the near future (and maybe it can work with v3 of the AWS SDK instead of v2).

lsp/server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -0,0 +1,183 @@
...
+                rightFileContent: right,
+            },
+            maxResults: params.maxResults,
+        }
+
+        const results: CodeWhispererClient.Recommendation[] = []
+
+        // We will get all the paginated recommendations.
+        // This is slow, and holds up the IDE's autocompletion list from showing.
+        // We wouldn't do this in a release.
```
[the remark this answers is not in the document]
awschristou: For a proof of concept, this is more of a brute force approach. The discussions focus around if/how IDEs support partial result sets, which we can accommodate in the server (the maxResults param is a nod to this).

lsp/server/aws-lsp-codewhisperer/src/client/codewhispererclient.d.ts
awschristou: I used the AWS Toolkit for VS Code repo to produce this client. I didn't set anything up in this repo (its not relevant for this proof of life phase).

PR #539, opened 2023-07-28
Update CodeWhisperer server to use bearer tokens instead of IAM credentials

awschristou:
This change demonstrates how we can let the IDE extension resolve and provide bearer tokens into language servers. This way, language servers do not have to maintain an identical bearer token implementation, which can drift over time. This change is the bearer token counterpart to #523, which worked with IAM based credentials.

When the server is started with a specific command line argument, the server waits to receive the encryption key over stdin before listening for the LSP protocol. Extensions then pass a marker through the LSP Initialization message. When the server detects this marker, it registers custom notification handlers that will be used to receive credentials and bearer tokens. The extension then pushes credentials and bearer tokens (using custom notification messages) whenever the credential state changes. In this way, the server always has credentials/tokens available for use. At times when no credentials/tokens can be resolved, the server's credentials/tokens are "un-set".

The sample CodeWhisperer language server has been updated to use bearer tokens. In the future, we will need a language server that can use either the IAM credentials or bearer tokens with CodeWhisperer. I have left the IAM credentials related code commented out (in lsp/server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts). It is left as a future exercise to make the language server support both modes based on initialization.

The sample VS Code extension has been updated with a command to obtain a bearer token, then push it to the language server. Much of this code was copied from the AWS Toolkit for VS Code repo. The following files are largely copied from that repo, and can be skipped while reviewing this change:
- lsp/client/vscode/src/sso/builderId.ts
- lsp/client/vscode/src/sso/model.ts
- lsp/client/vscode/src/tsUtils.ts

This concept is subject to additional changes before its initial release.

This change also deletes `lsp/client/vscode/package-lock.json` which addresses the repo's `npm audit` warnings. This should be okay because the sample vscode client is a part of the "monorepo" configuration, and uses `lsp/package-lock.json`

## License


lsp/client/vscode/src/sso/model.ts
```diff
@@ -0,0 +1,111 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// For the Proof of concept, this file's code was copied from the AWS Toolkit for VS Code repo
```
justinmk3, line 6: mentioning the commit sha (or better, a github url with the commit sha) helps a lot

awschristou: Good idea, I'll add these

awschristou:
> @awschristou is this pr still relevant ?

No, this code has moved to a separate repo, but we haven't removed it from this repo yet.
I will close this PR. We'll still be able to reference the closed PR for bearer token changes if needed.

PR #553, opened 2023-08-11
Concept: Update credentials transmission to use JWT

awschristou:
This change updates the proof of concept introduced in #523 illustrating how we can let the IDE extension handle credentials configuration/resolution, and provide credentials into the language servers. This change uses encrypted JSON Web Tokens (JWT) to transmit the credentials. This removes the need to explicitly handle things like data encryption or transmitting ivs and authTags.

## License


justinmk3 (approved):
Should this be called out in the guidance for other client impls?

awschristou:
> Should this be called out in the guidance for other client impls?

We're still determining what the right approach is for browser based hosts. We suspect this will be compatible, but will provide guidance once we have more certainty.

PR #586, opened 2023-10-20
feat(telemetry): support codewhispererterminal

chaynabors:
## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -855,6 +855,51 @@
             "name": "userChoice",
             "type": "string",
             "description": "User selection from a predefined menu (not user-provided input)" 
+        },
+        {
+            "name": "codewhispererterminal_command",
```
dhruvigajjar, line 860: we practice namespace_camelCaseMetricName. `codewhispererterminal_command` be renamed to `codewhisperer_terminalCommand`. my comment applies to all metric names
https://github.com/aws/aws-toolkit-common/blob/main/telemetry/telemetryformat.md#metrics

telemetry/definitions/commonDefinitions.json
```diff
@@ -855,6 +855,51 @@
...
+            "type": "string",
+            "description": "The CLI tool a completion was for"
+        },
+        {
+            "name": "codewhispererterminal_duration",
+            "type": "int",
+            "description": "A duration in milliseconds"
+        },
+        {
+            "name": "codewhispererterminal_duration",
```
dhruvigajjar, line 870: looks like `codewhispererterminal_duration` is redudant

telemetry/definitions/commonDefinitions.json
chaynabors: In our case, and as awful as it is, our intention was actually that the namespace _is_ codewhispererterminal

telemetry/definitions/commonDefinitions.json
chaynabors: The duration type in our language represents milliseconds using (I think) 128 bit unsigned ints. The top level duration is a double. This shouldn't matter, but does matter in our generated code.

dhruvigajjar (commented):
can you briefly describe the problem here. attach quip, targeted UI if you have to better related metrics' design.

telemetry/definitions/commonDefinitions.json
```diff
@@ -855,6 +855,51 @@
...
+            "description": "A count of characters typed"
+        },
+        {
+            "name": "codewhispererterminal_suggestedCount",
+            "type": "int",
+            "description": "A count of characters accepted"
+        },
+        {
+            "name": "codewhispererterminal_itemName",
+            "type": "string",
```
dhruvigajjar, line 896: just 'item' is too abstract in description. it would be helpful to know what do we consider as item.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3145,6 +3190,80 @@
                 { "type": "codecatalyst_devEnvironmentWorkflowStep"},
                 { "type": "codecatalyst_devEnvironmentWorkflowError", "required": false }
             ]
+        },
+        {
+            "name": "codewhispererterminal_userLoggedIn",
+            "description": "Emitted when users log in",
```
dhruvigajjar, line 3196: is this only emitted when user successfully logs in? do we have any metrics if user attempted to log in? you can return existing telemetry metrics too. 
e.g. there's credentialModification 
 {
            "name": "credentialModification",
            "type": "string",
            "allowedValues": ["Add", "Edit", "Delete"],
            "description": "The type of modification performed on the credentials"
        }

telemetry/definitions/commonDefinitions.json
```diff
@@ -3145,6 +3190,80 @@
...
+            "passive": false,
+            "metadata": []
+        },
+        {
+            "name": "codewhispererterminal_completionInserted",
+            "description": "Emitted on Autocomplete user decisions",
+            "passive": false,
+            "metadata": [
```
dhruvigajjar, line 3204: are all `codewhispererterminal_completionInserted`  `passive:false` ?

dhruvigajjar: will this metric be used in calculating user success rate? if yes then consider adding result field.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3145,6 +3190,80 @@
...
+            "name": "codewhispererterminal_dashboardPageViewed",
+            "description": "Emitted when CW CLI user fails a doctor check",
+            "passive": false,
+            "metadata": [
+                { "type": "codewhispererterminal_route" }
+            ]
+        },
+        {
+            "name": "codewhispererterminal_menuBarClicked",
+            "description": "Emitted when a user clicks the menu bar",
```
dhruvigajjar, line 3256: you could have a click count as discussed in slack. collapsing it to have number of counts for a duration or session

telemetry/definitions/commonDefinitions.json
```diff
@@ -3145,6 +3190,80 @@
...
+            "passive": false,
+            "metadata": []
+        },
+        {
+            "name": "codewhispererterminal_menuBarItemClicked",
+            "description": "Emitted when a user clicks an item in the menu bar",
+            "passive": false,
```
dhruvigajjar, line 3263: same comment as codewhispererterminal_menuBarClicked

telemetry/definitions/commonDefinitions.json
chaynabors: nope, oversight

telemetry/definitions/commonDefinitions.json
mschrage: I think it is okay to camelCase this? (Don't feel strongly either way but would prefer to stick with convention)

dhruvigajjar: I think this is first time we have seen two word namespace. but let it be `codewhispererterminal` since camelCase get's converted to `codewhisperer_terminal_command` which is wrong as this format suggests `codewhisperer` metric and NOT `codewhispererterminal`'s metric.

PR #596, opened 2023-10-27
Bump version to unblock pipeline

jpinkney-aws:
## Problem
Build failed due to not being able to update version.

## Solution
Manually update the build version

## License


dianaabv16 (approved):

nkomonen-amazon:
The buildspec (`AWSToolkitsPackagingInfrastructure` `cloneRepo.yml`) in our internal repo has a condition that if only the `version` file is updated it will not trigger a new build. So this PR does not work.

PR #600, opened 2023-10-30
Bump version to unblock pipeline 

dianaabv16:
## Problem
force update the version due to build issue 
- We need to make a change to trigger a new release that is not only updating the 'version' file

## License


vishaalmehrishi (approved):

jpinkney-aws (approved):

justinmk3:
This was presumably caused by the NPM outage: [There is an NPM outage. Did it get resolved? If not, this presumably won't work.](https://status.npmjs.org/incidents/zdznxkrp22py)

Looks like that was resolved so we can try this.

PR #617, opened 2023-11-10
feat(tele): Add amazonq_ (awqs_) metrics

dianaabv16:
## Problem
- currently the name of telemetry metrics for weaverbird is `awsq_` it should be `amazonq_`
- we are changing the format of out metrics.  we want to able to have more centralised metrics.

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -2663,6 +2704,89 @@
             "description": "User has accepted the approach generated by Weaverbird.",
             "metadata": [{ "type": "enabled" }, { "type": "awsqConversationId" }] 
         },
+        {
```
jpinkney-aws, line 2707: I see that a few things are added but none are deleted, are we going to remove all the awsq ones that are currently being used here or is that a next step?

dianaabv16: I think to make it backwards compatible.
I think it is better to keep awsq_ metrics until we are fully moved to aamazonq_
what do you think>

jpinkney-aws: I think that's fine so we don't block other teams but we have to have a task to make sure to remove them

dianaabv16: I will add one

dhruvigajjar: Is the change live?

jpinkney-aws (approved):

telemetry/definitions/commonDefinitions.json
dianaabv16: no, I this is to register new events.I will open second CR in aws-toolkit-vscode-staging

jpinkney-aws: Yes, we've had awsq metrics for ~1.5 weeks

PR #622, opened 2023-11-17
feat(amazonq): add missing metrics

dianaabv16:
## Problem
- remove amazonq_approachIteration and amazonq_codeGenerationIteration.
- moved its metrics to  amazonq_approachInvoke and amazonq_codeGenerationInvoke accordingly
- added amazonqEndOfTheConversationLatency and amazonqNumberOfReferences
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2651,6 +2651,11 @@
             "metadata": [{ "type": "enabled" }, { "type": "amazonqConversationId" }]
         },
         {
+            "name": "amazonq_isProvideFeedbackForCodeGen",
+            "description": "User asked to regenerate code generation with a comment",
+            "metadata": [{ "type": "enabled" }, { "type": "amazonqConversationId" }]
+        },
+        {
             "name": "amazonq_isReviewedChanges",
             "description": "User reviewed changes",
             "passive": true,
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -901,11 +901,20 @@
             "type": "double",
             "description": "Captures the number of files generated as a part of code generation iteration"
         },
-
         {
             "name": "amazonqRepositorySize",
             "type": "double",
             "description": "Captures  size of the source code"
+        },
+        {
+            "name": "amazonqNumberOfReferences",
+            "type": "double",
+            "description": "Captures the number of references"
+        },
+        {
+            "name": "amazonqEndOfTheConversationLatency",
+            "type": "double",
+            "description": "The time it takes to end amazonq conversation"
         }
     ],
     "metrics": [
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -901,11 +901,20 @@
...
+        },
+        {
+            "name": "amazonqNumberOfReferences",
+            "type": "double",
+            "description": "Captures the number of references"
+        },
+        {
+            "name": "amazonqEndOfTheConversationLatency",
+            "type": "double",
+            "description": "The time it takes to end amazonq conversation"
```
jpinkney-aws, line 917: This is technically the entire duration of the conversation right? Can we say something like the total time from start to end

telemetry/definitions/commonDefinitions.json
```diff
@@ -2602,62 +2611,53 @@
...
-            "name": "amazonq_approachIteration",
-            "description": "User recieved a result in Approach Iteration stage",
+            "name": "amazonq_approachInvoke",
+            "description": "Captures Approach generation process",
             "metadata": [
                 { "type": "amazonqConversationId" },
                 { "type": "amazonqGenerateApproachIteration" },
-                { "type": "amazonqGenerateApproachLatency" }
+                { "type": "amazonqGenerateApproachLatency" },
+                { "type": "result" },
```
jpinkney-aws, line 2628: Do we need to explicitly add these? I thought they came from the metric base

telemetry/definitions/commonDefinitions.json
```diff
@@ -2602,62 +2611,53 @@
...
                 { "type": "amazonqConversationId" },
                 { "type": "amazonqGenerateCodeIteration" },
                 { "type": "amazonqGenerateCodeResponseLatency" },
                 { "type": "amazonqCodeGenerationResult" },
-                { "type": "amazonqNumberOfFilesGenerated" },
-                { "type": "amazonqRepositorySize" }
+                { "type": "amazonqNumberOfFilesGenerated", "required": false },
+                { "type": "amazonqRepositorySize" },
+                { "type": "amazonqNumberOfReferences", "required": false },
+                { "type": "result" },
```
jpinkney-aws, line 2648: Same here, I don't think result/reason are strictly required. IIRC we get them automatically

dianaabv16: let's keep for readability purposes

telemetry/definitions/commonDefinitions.json
dianaabv16: let's keep for readability purposes

telemetry/definitions/commonDefinitions.json
dianaabv16: renamed

telemetry/definitions/commonDefinitions.json
jpinkney-aws: I think we should still remove it because it goes against the standards of the project (AFAICT no other teams are doing this)

nkomonen-amazon (approved):

telemetry/definitions/commonDefinitions.json
nkomonen-amazon: All metrics have result/reason (see `MetricBase` interface) so you will get autocompletion for result/reason. I think the goal is to keep these minimal so we don't need to add `result` since it is implicit, I will look to update docs somewhere to indicate this

telemetry/definitions/commonDefinitions.json
```diff
@@ -2602,62 +2611,53 @@
...
+                { "type": "reason", "required": false }
             ]
         },
         {
             "name": "amazonq_isAcceptedCodeChanges",
             "description": "User applied code changes generated for the task.",
             "metadata": [{ "type": "enabled" }, { "type": "amazonqConversationId" }]
         },
-        {
-            "name": "amazonq_isRejectedCodeChanges",
```
jpinkney-aws, line 2654: We probably want to switch this to amazonq_isProvideFeedbackForCodeGen ? Since that's still an option

dianaabv16: added

PR #641, opened 2023-12-06
Add unit tests to common package for allowedValues change

damntrecky:
## Problem
We rushed out a fix yesterday because there were `allowedValues` with `spaces in the values` and this wasn't supported by the build here. We didn't add unit tests for the fix to convert spaces to `_` underscores

## Solution
Unfortunately it looks like for Kotlin and C# the `allowedValues` enum, or output metdata key WAS converted, but the value itself was NOT converted to have `_` underscores.

This PR fixes that and adds unit tests for all three languages. I also updated the README for each section as it might not be obvious for developers on how to run the tests for each language. Especially if they are not familiar with that language.
## License


telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
```diff
@@ -213,7 +213,7 @@ private CodeTypeDeclaration GenerateEnumStruct(MetricType type)
                     allowedValue.Replace(" ", "_").ToPascalCase().Replace(".", "").Replace("-", ""))
                 {
                     InitExpression = new CodeObjectCreateExpression(type.GetGeneratedTypeName(),
```
damntrecky, line 215: For C# we converted the "key" but not the "value" to get rid of spaces. This was an oversight and testing of course caught this. 

Thanks for pushing for testing.

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/GeneratedCodeTests.cs
```diff
@@ -217,6 +217,40 @@ public void RecordMetricWithNewTransform()
             Assert.Single(datum.Metadata);
         }
```
damntrecky, line 219: Testing the actual record that introduced the issue with "spaces" in the `allowedValues`

telemetry/jetbrains/src/main/kotlin/software/aws/toolkits/telemetry/generator/TelemetryGenerator.kt
```diff
@@ -86,7 +86,7 @@ private fun FileSpec.Builder.generateTelemetryEnumType(item: TelemetryMetricType
     item.allowedValues!!.forEach { enumValue ->
         enum.addEnumConstant(
             enumValue.toString().replace(Regex("\\s"), "_").toTypeFormat(), TypeSpec.anonymousClassBuilder()
```
damntrecky, line 88: For Kotlin we converted the "key" but not the "value" to get rid of spaces. This was an oversight and testing of course caught this. 

Thanks for pushing for testing.

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
awschristou: Were any Toolkits released with spaces in the values? This change would create a discrepancy in emitted values across Toolkit versions.

Why not change the definition data to remove the spaces? I don't understand why the values cannot contain spaces.

telemetry/jetbrains/src/test/resources/testGeneratorAllowedValuesOutput
```diff
@@ -0,0 +1,101 @@
...
+import software.aws.toolkits.jetbrains.services.telemetry.MetricEventMetadata
+import software.aws.toolkits.jetbrains.services.telemetry.TelemetryService
+
+/**
+ * A test object for parsing allowedValues
+ */
+public enum class TestAllowedValues(
+    private val `value`: String,
+) {
+    TestSpacesAreReplaced("test spaces are replaced"),
```
damntrecky, line 25: Key is converted, but NOT the value. In general the advice was to avoid spaces for values.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1041,14 +1041,14 @@
...
             "description": "Names of the pre-validation errors that can occur in the project",
             "allowedValues": [
-                "No pom.xml file found",
-                "No Java project found",
-                "Mixed Java project and another language found",
-                "Project selected is not Java 8 or Java 11",
-                "Only Maven projects supported",
-                "Empty project",
-                "Non SSO login",
-                "Project running on backend"
```
damntrecky, line 1051: Update the values to NOT use spaces.

telemetry/vscode/src/generate.ts
```diff
@@ -81,10 +81,7 @@ function getArgsFromMetadata(m: MetadataType): string {
                 name,
                 kind: StructureKind.TypeAlias,
                 isExported: true,
-                type: m.allowedValues.map(v => {
-                    const modifiedVal = typeof v === "string" ? v.replace(/\s/g, "_") : v;
-                    return `'${modifiedVal}'`
-                }).join(' | '),
```
damntrecky, line 87: Revert conversion of values to _ values in VSCode.

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
damntrecky: There were spaces deployed this week. But we can revert this quickly as our slack discussion aligns with this being better for backend ingestion now and in the future.

telemetry/jetbrains/src/test/resources/testGeneratorAllowedValuesOutput
damntrecky: I would  expect this not be a breaking change, since no other `allowedValues` have spaces other than the values my team introduced this week.

damntrecky:
Kotlin build failing fixed here -> https://github.com/aws/aws-toolkit-common/pull/646/files

I believe this current change WILL be a breaking changes for anyone who pulls it in and should be separate from the change above. Ideally, once this change is deployed my team can fix the breaking changes (which are for Elastic gumby/ amazon q code transform) intelliJ and vscode instances.

awschristou:
@damntrecky this PR contains two separate changes:
- updating enum values in telemetry/definitions/commonDefinitions.json to preferred values (removing spaces from values)
- fixing and validating generators to handle spaces in values

These need to be split into two separate PRs. If we need to investigate a regression, or revert a change, this PR as-is makes the repo harder to manage. The enum values update is needed sooner than the validation update. Please create a separate PR with just that change. Once that is merged in, this PR can be revisited in order to handle the testing (which is less urgent, as we won't have enum values containing spaces).

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/GeneratedCodeTests.cs
```diff
@@ -217,6 +217,41 @@ public void RecordMetricWithNewTransform()
...
 
+        /// <summary>
```
damntrecky, line 220: Test to make sure `allowedValues` with spaces converts the key properly and make sure the allowedValue -> value has spaces for now.

Although its recommended to NOT have spaces in the values, the build should not fail and have consistent behavior across environments

telemetry/csharp/AwsToolkit.Telemetry.Events.Generator/DefinitionsBuilder.cs
justinmk3: should we validate the `allowedValues` so that CI fails if they contain spaces?

telemetry/csharp/AwsToolkit.Telemetry.Events.Tests/GeneratedCodeTests.cs
```diff
@@ -217,6 +217,41 @@ public void RecordMetricWithNewTransform()
...
+        /// RecordCodeTransformIsDoubleClickedToTriggerInvalidProject was chosen as a sample call that has a 
+        /// CodeTransformPreValidationError for allowedValues which have spaces in the values that
```
justinmk3, line 222: `CodeTransformPreValidationError` values no longer have spaces, does the comment need to be updated?

Frank97Tyler (approved):

PR #650, opened 2023-12-13
Missing Event for logging when users toggle telemetry on/off

damntrecky:
## Problem
We do not log when users turn off/on telemetry

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -1,5 +1,11 @@
 {
     "types": [
+        {
+            "name": "toolkitTelemetryStatusType",
+            "type": "string",
+            "description": "The type to determine the users status of having telemetry event submitted or not.",
+            "allowedValues": ["OFF", "ON"]
```
justinmk3, line 7: is a new type needed for this or could it use this: https://github.com/aws/aws-toolkit-common/blob/ffab9084e1d0056aed79eb65ac4dffa506eb9f7d/telemetry/definitions/commonDefinitions.json#L257-L260

damntrecky: If we do carry this out, I would like to re-use types. I will delete this type from the PR.

awschristou:
The telemetry system implemented into each of the three Toolkits does not support this use case. Those systems are deliberately designed to ensure that metrics are not transmitted when users opt-out of telemetry. I'm closing this PR, since there is no value in defining a metric that can't be used the way you've intended.

PR #665, opened 2024-01-08
added ComputeEnv as type with dev, local, test, web, cc, cloud9, wsl,…

dhruvigajjar:
## Problem
Problem : Identify the ENV in metrics. Since different env have different user flows, this field becomes helpful in understanding why some metrics are emitted or better understand user flow w.r.t to ENVS

initially adding "ec2", "codecatalyst", "cloud9", "wsl", "test", "vscode-web", "sagemaker", etc.
we can add more later
field name = computeEnv, analogous to the existing computeRegion field
## Solution
For all Toolkits to utilize field, adding ComputeEnv Field in PostMetricsRequest & PostFeedbackRequest

## License

### DIFF
--- a/telemetry/service/service-model.json
+++ b/telemetry/service/service-model.json
@@ -68,15 +68,17 @@
       "type":"string",
       "enum":[
         "cloud9",
-        "codeCatalyst",
-        "dev",
+        "cloud9-codecatalyst",
+        "cloud9-ec2",
+        "codecatalyst",
         "ec2",
         "local",
         "sagemaker",
         "ssh",
         "test",
         "web",
-        "wsl"
+        "wsl",
+        "other"
       ]
     },
     "Datapoint":{
--- a/telemetry/service/service-model.json
+++ b/telemetry/service/service-model.json
@@ -64,6 +64,21 @@
       "type":"string",
       "max":2000
     },
+    "CommputeEnv":{
+      "type":"string",
+      "enum":[
+        "cloud9",
+        "codeCatalyst",
+        "dev",
+        "ec2",
+        "local",
+        "sagemaker",
+        "ssh",
+        "test",
+        "web",
+        "wsl"
+      ]
+    },
     "Datapoint":{
       "type":"double",
       "min":0
@@ -169,6 +184,7 @@
         "AWSProductVersion":{"shape":"AWSProductVersion"},
         "OS":{"shape":"Value"},
         "OSVersion":{"shape":"Value"},
+        "CommputeEnv": {"shape":"ComputeEnv"},
         "ParentProduct":{"shape":"Value"},
         "ParentProductVersion":{"shape":"Value"},
         "Metadata":{"shape":"Metadata"},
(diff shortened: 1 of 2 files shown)


telemetry/service/service-model.json
```diff
@@ -64,6 +64,21 @@
       "type":"string",
       "max":2000
     },
+    "CommputeEnv":{
+      "type":"string",
+      "enum":[
+        "cloud9",
+        "codeCatalyst",
```
hayemaxi, line 71: do we need a type for a Cloud9 + CodeCatalyst?

telemetry/service/service-model.json
```diff
@@ -64,6 +64,21 @@
       "type":"string",
       "max":2000
     },
+    "CommputeEnv":{
```
justinmk3, line 67:
```suggestion
    "ComputeEnv":{
```

telemetry/service/service-model.json
justinmk3: since this is (currently) an enum instead of a list, yes. there should be a `cloud9Codecatalyst` item

telemetry/service/service-model.json
```diff
@@ -64,6 +64,21 @@
...
+      "type":"string",
+      "enum":[
+        "cloud9",
+        "codeCatalyst",
+        "dev",
```
justinmk3, line 72: is this for "gamma" or something like that?

telemetry/service/service-model.json
rli, line 71: why not `codecatalyst`?

telemetry/service/service-model.json
rli, line 1: i do not remember the exact context why this file is in the public repo, but changes need to be backported to the internal `.xml` file to take effect service-side

telemetry/service/service-model.json
dhruvigajjar: what are other combinations? if env can be combined & there are multiple combinations, it is good idea to instrument on Tookits' code insead?

justinmk3: > what are other combinations? if env can be combined

That's a judgment call. The idea here (and in the quip doc) is to decide out. I think we have decided it by now. The combinations are as we've listed them here (after making the changes mentioned in this PR).

dhruvigajjar: got it.

telemetry/service/service-model.json
dhruvigajjar: actually i added two words for dev & local that mean the same. removed it

telemetry/service/service-model.json
dhruvigajjar: Done

telemetry/service/service-model.json
dhruvigajjar: https://code.amazon.com/reviews/CR-111231197/revisions/1#/details

hayemaxi (approved):

justinmk3 (approved):

PR #666, opened 2024-01-11
fix: missing dependabot updates

justinmk3:
# Problem:
Since ebe7d3cea52169c4f3ebdc398517c4a1a469e80d, dependabot does not send updates for packages not matching a group. For example, in the "telemetry/vscode/" directory it will only update `jest` and `@types` dependencies. Other deps won't be updated.

Hint: https://github.com/aws/aws-toolkit-common/pull/533#issuecomment-1886793932

# Solution:
- Review the inscrutable documentation: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
- Reference examples on github via: https://github.com/search?q=path%3Adependabot.yml+groups+all&type=code
- Remove `dependency-type` directive.
- Try `allow: dependency-name: '*'` directive. 
    - If this doesn't work, later try a "catch-all" group. Note that `groups` are evaluated as "first wins".
- Note: intentionally _not_ using ["allow: dependency-type:all" directive](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow); it enables updates for _transitive_ dependencies.

## License

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,38 +1,48 @@
-# Please see the documentation for all configuration options:
-# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+# Documentation for configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 # https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates
 
 version: 2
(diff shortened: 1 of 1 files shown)


justinmk3:
There is no way to validate dependabot configs without commiting, so just going to yolo this.

justinmk3:
config is valid: https://github.com/aws/aws-toolkit-common/runs/20382074101

now need to wait ~1 day...

PR #677, opened 2024-01-23
fix(gumby): modify metric to track errors

dhasani23:
## Problem

We run a `mvn` and `javap` shell command and want to track the error message that results when these commands fail.

## Solution

Add a field to the relevant metrics to keep track of the error message.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -3541,6 +3541,14 @@
                 {
                     "type": "codeTransformRuntimeError",
                     "required": false
+                },
+                { 
+                    "type": "result", 
+                    "required": true 
+                },
+                { 
+                    "type": "reason", 
+                    "required": true 
                 }
             ]
         },
@@ -3894,6 +3902,14 @@
                 {
                     "type": "codeTransformRuntimeError",
                     "required": true
+                },
+                { 
+                    "type": "result", 
+                    "required": true 
+                },
+                { 
+                    "type": "reason", 
+                    "required": true 
                 }
             ]
         }
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -3538,8 +3538,10 @@
                     "type": "codeTransformSessionId",
                     "required": true
                 },
-                { "type": "result", "required": true },
-                { "type": "reason", "required": true }
+                {
+                    "type": "codeTransformRuntimeError",
+                    "required": false
+                }
             ]
         },
         {
@@ -3890,8 +3892,8 @@
                     "required": true
                 },
                 {
-                    "type": "reason",
-                    "required": false
+                    "type": "codeTransformRuntimeError",
+                    "required": true
                 }
             ]
         }


telemetry/definitions/commonDefinitions.json
```diff
@@ -3538,8 +3538,10 @@
                     "type": "codeTransformSessionId",
                     "required": true
                 },
-                { "type": "result", "required": true },
-                { "type": "reason", "required": true }
```
dhasani23, line 3542: We do emit result and reason for this metric, and I have no plan to change this, but I removed these here because I noticed that they are missing in many other metrics, so I want to be consistent.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3890,8 +3892,8 @@
                     "required": true
                 },
                 {
-                    "type": "reason",
-                    "required": false
+                    "type": "codeTransformRuntimeError",
+                    "required": true
```
dhasani23, line 3896: Required here because when `mvn` command fails we should always have an error message to emit.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3538,8 +3538,10 @@
                     "type": "codeTransformSessionId",
                     "required": true
                 },
-                { "type": "result", "required": true },
-                { "type": "reason", "required": true }
+                {
+                    "type": "codeTransformRuntimeError",
+                    "required": false
```
dhasani23, line 3543: Not required here because we use this `codeTransform_isDoubleClickedToTriggerInvalidProject` in many spots, in some of which it is not necessary to include any type of error message.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3890,8 +3892,8 @@
                     "required": true
                 },
                 {
-                    "type": "reason",
-                    "required": false
```
dhasani23, line 3894: See first comment at very top.

telemetry/definitions/commonDefinitions.json
dhasani23: Note that `codeTransformRuntimeError` is an existing field we have (already defined as a string above)

JuliaABurch (approved):

telemetry/definitions/commonDefinitions.json
damntrecky: I think we are trying to keep `result` and `reason` as part of our new standard. What is the goal here from removing this?

tincheng:
Just discussed with AWS Toolkit team on metric improvement, and moving forward we want to avoid using code transform specific fields, but since codeTransfromRuntimeError is an existing field, to align with other current error metrics, we can keep using that fields. We don't want to remove error/reason fields.

tincheng (approved):

telemetry/definitions/commonDefinitions.json
dhasani23: Added them back.

telemetry/definitions/commonDefinitions.json
dhasani23: EDIT: Added them back.

tincheng: Moving forward we will want to have reason/result in all of our metrics, but since we are already working on revamping the metric schema and the majority of these metrics will be replaced, I think it's fine to not back-fill the fields at the moment.

telemetry/definitions/commonDefinitions.json
nkomonen-amazon: The `reason` field should be sufficient to tell you why it failed. Is there a reason you specifically need this field? Can the value that would go in to this field be put in to `reason` instead?

telemetry/definitions/commonDefinitions.json
nkomonen-amazon: Same as my other comment above

telemetry/definitions/commonDefinitions.json
damntrecky: If IntelliJ uses the `codeTransformRuntimeError` field already then we should use this field. If not, we should move the error into the `reason` field if possible.

telemetry/definitions/commonDefinitions.json
damntrecky: If IntelliJ uses the `codeTransformRuntimeError` field already then we should use this field. If not, we should move the error into the `reason` field if possible.

telemetry/definitions/commonDefinitions.json
dhruvigajjar: Not just relate to this case
if any field is required true, that means it needs to be populated. if you are concerned about consistency, required:true is nice way to force it.

telemetry/definitions/commonDefinitions.json
dhruvigajjar: What kind of error message is expected here? Do you have any sample format. 
Telemetry metrics are for overall analysis and we don't want to have long texts. maybe a gist in reason is good practice. 

if you need long error messages, we can dump it to cloudwatch logs

dhasani23:
Looks like we should be logging to CW Logs for longer error messages, so will close this PR

PR #681, opened 2024-01-30
Nardeck/maven telemetry changes

damntrecky:
## Problem

Logging for maven build failures is a bit under thought. We want to introduce some slight changes that we think will help normalize logging these failures for Telemetry/BI.

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -1078,14 +1078,24 @@
             ]
         },
         {
-            "name": "codeTransformMavenBuildCommand",
+            "name": "codeTransformCommandExecutables",
```
dhasani23, line 1081: Maybe rename to `codeTransformBuildToolExecutable`

telemetry/definitions/commonDefinitions.json
```diff
@@ -3904,12 +3922,20 @@
                     "required": true
                 },
                 {
-                    "type": "codeTransformMavenBuildCommand",
+                    "type": "codeTransformCommandExecutables",
+                    "required": true
+                },
+                {
+                    "type": "codeTransformRuntimeError",
```
dhasani23, line 3929: We already put the failure reason under `reason` -- do we want this so that we can hard-code the `reason` field?

telemetry/definitions/commonDefinitions.json
```diff
@@ -3556,6 +3566,14 @@
                     "type": "codeTransformSessionId",
                     "required": true
                 },
+                {
+                    "type": "codeTransformRuntimeError",
+                    "required": true
```
dhasani23, line 3571: Should be `false` because it's not always applicable. Either way this was merged recently in my PR.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3556,6 +3566,14 @@
...
+                },
+                {
+                    "type": "codeTransformCommandNames",
```
dhasani23, line 3574: Why do we need this field here? I don't think it applies to this metric, besides for `javap`, but we already handle that: `codeTransformPreValidationError: 'NoJavaProject'`

telemetry/definitions/commonDefinitions.json
```diff
@@ -1078,14 +1078,24 @@
...
                 "mvn",
                 "mvnw",
                 "IDEBundledMaven"
             ]
+        },
+        {
+            "name": "codeTransformCommandNames",
+            "type": "string",
+            "description": "Type of maven, gradle and other command line arguments run in the IDE",
+            "allowedValues": [
```
dhasani23, line 1094: I do agree that we need something like this for the shell commands

telemetry/definitions/commonDefinitions.json
```diff
@@ -1078,14 +1078,24 @@
             ]
         },
         {
-            "name": "codeTransformMavenBuildCommand",
+            "name": "codeTransformCommandExecutables",
             "type": "string",
             "description": "Type of maven command",
             "allowedValues": [
                 "mvn",
                 "mvnw",
```
dhasani23, line 1086: Potentially add `mvn.cmd` and `mvnw.cmd`? Or a boolean field for Windows?

telemetry/definitions/commonDefinitions.json
damntrecky: I'm trying to get a more generic name as we might not only add build commands here. Could be other setup commands for other build tools and looking ahead other programs.

telemetry/definitions/commonDefinitions.json
damntrecky: Lets exhaust all the options we want in here offline.

telemetry/definitions/commonDefinitions.json
damntrecky: My thought: There are sub groups within the preValidationError. For example 'NoJavaProject' can have two different paths to trigger in VSCode right?

telemetry/definitions/commonDefinitions.json
damntrecky: My thoughts: We will use the `reason` field to have `MavenInstall` or `MavenCopy` and the runtime error will be the output of the command.

telemetry/definitions/commonDefinitions.json
damntrecky: Noted: should make sure this is updated to false during merge/rebase

telemetry/definitions/commonDefinitions.json
```diff
@@ -3530,6 +3540,26 @@
                 }
             ]
         },
+        {
```
damntrecky, line 3543: We need to add an event to trigger for when projects have not yet initiated transformation to get user details. This is a P0 requirement.

PR #683, opened 2024-01-31
feature(QTransform): Add telemetry for 1p dependencies

carolabadeer:
## Problem
A large amount of applications submitted through Amazon Q Code Transformation fail at the pre-build stage due to errors in retrieving dependencies, especially when dependencies not available on maven central (1p dependencies) are not able to be fetched and packaged up by Q Transform. This metric gives us visibility into whether a project has dependencies that cannot be found on maven central, and the data will be used (1) for the team to get a better understanding of the percentage of pre-build failures associated with 1p dependencies, and (2) to design and implement future enhancements in the case of Q Transform being unable to fetch certain private dependencies. 

## Solution
Add a new true/false metric for the existence of 1p dependencies in an application. See associated PRs in VSCode and IntelliJ: 
- [VSCode](https://github.com/aws/aws-toolkit-vscode/pull/4308) 
- [IntelliJ](https://github.com/aws/aws-toolkit-jetbrains/pull/4109)

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1076,6 +1076,11 @@
                 "mvnw",
                 "IDEBundledMaven"
             ]
+        },
+        {
+            "name": "codeTransformHas1pDependency",
+            "type": "boolean",
+            "description": "Project contains at least 1 dependency not available in maven central"
         }
     ],
     "metrics": [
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -3894,6 +3899,24 @@
                     "required": false
                 }
             ]
+        },
+        {
+            "name": "codeTransform_projectContains1pDependencies",
```
justinmk3, line 3904: Is a new metric needed for this, or would it make sense as a flag on an existing metric? 

Metrics should map to user actions, or significant events. There should not be a new metric every time we have a new bit of data to collect.

justinmk3:
"This PR can be closed, we did this implementation in the QCT backend"

PR #690, opened 2024-02-07
feat: Add amazonqUploadIntent field to amazonq_createUpload

jsalles:
## Problem

We have a missing field in our amazonq_createUpload capture, which identifies the intent of the upload.

## Solution

This CR adds that field as optional with possible values of `TRANSFORMATION` or `TASK_ASSIST_PLANNING`.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -929,6 +929,12 @@
             "description": "Captures the size of the source code"
         },
         {
+            "name": "amazonqUploadIntent",
+            "type": "string",
+            "description": "The intent of the upload",
+            "allowedValues": ["TRANSFORMATION", "TASK_ASSIST_PLANNING"]
+        },
+        {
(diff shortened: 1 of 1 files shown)


dianaabv16 (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -2816,7 +2822,8 @@
             "description": "Captures createUploadUrl invocation process",
             "metadata": [
                 { "type": "amazonqConversationId" },
-                { "type": "amazonqRepositorySize", "required": false }
+                { "type": "amazonqRepositorySize", "required": false },
+                { "type": "amazonqUploadIntent", "required": false }
```
vishaalmehrishi, line 2826: Given that this is specifically for `amazonq_createUpload`, should it be required? There are two allowed values and ideally each telemetry emission should be for either one upload intent.

nkomonen-amazon (approved):

PR #691, opened 2024-02-08
telemetry(CodeTransform): add status field to vcs metrics

dhasani23:
## Problem

We want to be able to know, when a user submits or cancels our proposed changes, was the project status successful or partially successful (at this point in the transformation, it must be one of these two).

## Solution

Add existing status field to two existing metrics.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1074,10 +1074,10 @@
             "description": "Names of the pre-validation errors that can occur",
             "allowedValues": [
                 "NoPom",
-                "NoJavaProjects",
+                "NoJavaProject",
                 "MixedLanguages",
                 "UnsupportedJavaVersion",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -3797,6 +3797,10 @@
                 {
                     "type": "codeTransformJobId",
                     "required": false
+                },
```
dhruvigajjar, line 3800: if this is `required:true` all existing implementation will need update to populate this field.

dhruvigajjar (approved):

telemetry/definitions/commonDefinitions.json
dhasani23: That's fine. We want to update them to include this field.

damntrecky (approved):

dhruvigajjar (approved):

PR #699, opened 2024-02-16
Measure whether CodeWhisperer is enabled or not

awschristou:
## Problem

We want to understand how often users turn CodeWhisperer on or off.

## Solution

Measure the current state at startup, and when explicitly turned on/off.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -3450,6 +3450,21 @@
             ]
         },
         {
+            "name": "codewhisperer_enabled",
+            "description": "Whether or not CodeWhisperer is enabled",
+            "passive": true,
+            "metadata": [
+                {
+                    "type": "source",
+                    "required": true
+                },
+                {
+                    "type": "enabled",
+                    "required": true
+                }
+            ]
+        },
+        {
             "name": "codecatalyst_createDevEnvironment",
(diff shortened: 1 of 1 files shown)


dhruvigajjar (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -3449,6 +3449,21 @@
                 }
             ]
         },
+        {
+            "name": "codewhisperer_enabled",
+            "description": "Whether or not CodeWhisperer is enabled",
+            "passive": true,
```
shruti0085, line 3455: Shouldn't this be false for when state change occurs? Users would have explicitly toggled this?

justinmk3: Yeah this doesn't seem passive. Also could use the existing metric `aws_modifySetting` (which is not passive): https://github.com/aws/aws-toolkit-common/blob/668213ff14128fbb1b4ab846ab0841e086e93b2c/telemetry/definitions/commonDefinitions.json#L2755C22-L2755C39

awschristou: Because this event can occur at arbitrary times (like startup), it should never be set to active, as this would generate false actives. This metric isn't critical in measuring active use (there are other actions/metrics that do this), so having perfect "active use" fidelity around this metric would not give us any additional insights.

PR #701, opened 2024-02-19
feat(amazonq): Add credential start url for amazonq events

leigaol:
## Problem
Amazon Q telemetry events should have startUrl in its field.  Because there is a need for more granularity in user connections in order to help diagnose problems.

## Solution
Add credential start url for amazonq events. Option for backward compatibility.

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -2791,7 +2791,8 @@
             "description": "Captures end of the conversation with amazonq /dev",
             "metadata": [
                 { "type": "amazonqConversationId" },
-                { "type": "amazonqEndOfTheConversationLatency", "required": false }
+                { "type": "amazonqEndOfTheConversationLatency", "required": false },
+                { "type": "credentialStartUrl", "required": false }
             ]
         },
```
dhruvigajjar, line 2797: will all amazon Q metrics will have `credentialStartUrl`? or is it like after auth, all metrics will have.
In which cases `credentialStartUrl` will be empty?

dhruvigajjar: so for users who's auth is expired, we wont have startUrl or we will have last valid startUrl?

leigaol: 1. It is the current active start url.
2. If auth expired, there won't be any startUrl. User with expired auth won't be able to invoke the Code path in Q that sends out any telemetry. Without a valid auth, they won't be able to interact with this product at all. Without any valid user interaction, there won't be any telemetry data.

justinmk3:
merge conflict

leigaol:
> merge conflict

Fixed in latest commit

telemetry/definitions/commonDefinitions.json
dhruvigajjar: So which metrics do you emit without auth? 
Does that mean some metric must have startUrl because they wont be emitted if auth is invalid?

hayemaxi (approved):

PR #703, opened 2024-02-28
Gumby - Add codeTransform_jobMavenRunTime to track maven commands execution

mk-fan:
## Problem
- Need better telemetry to track Gumby's local Maven commands execution

## Solution
- Add codeTransform_jobMavenRunTime

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -982,6 +982,21 @@
             "type": "string",
             "description": "The API request ID"
         },
+        {
+            "name": "codeTransformMavenCommands",
+            "type": "string",
+            "description": "Maven commands that have been run."
+        },
+        {
+            "name": "codeTransformMavenExecutionSuccessful",
```
dhasani23, line 991: They won't be able to even upload the project / start the job in either IDE if any Maven command fails

telemetry/definitions/commonDefinitions.json
```diff
@@ -982,6 +982,21 @@
             "type": "string",
             "description": "The API request ID"
         },
+        {
+            "name": "codeTransformMavenCommands",
+            "type": "string",
+            "description": "Maven commands that have been run."
```
dhasani23, line 988: We already have a field that does mostly the same thing: `CodeTransformMavenBuildCommand`. And we do emit telemetry that will tell us the specific command that failed (`install` or `copy-dependencies`).

telemetry/definitions/commonDefinitions.json
```diff
@@ -982,6 +982,21 @@
...
+        },
+        {
+            "name": "codeTransformMavenExecutionSuccessful",
+            "type": "boolean",
+            "description": "If the Maven commands executions are successful."
```
justinmk3, line 993: There is a standard `result` field on every metric. That should strongly be preferred to represent success/fail/cancel

telemetry/definitions/commonDefinitions.json
```diff
@@ -3808,6 +3823,28 @@
                 }
             ]
         },
+        {
+            "name": "codeTransform_jobMavenRunTime",
+            "description": "A timestamp for when all maven commands are done executing",
```
justinmk3, line 3828: A timestamp is not an event. It's a field. Is there an existing event that can capture this? You can add fields to an existing event.

mk-fan: Currently codeTransform_jobCreateZipEndTime captures the time of maven commands execution and the time of zip file creation combined. We want to track them separately. The only related existing event is codeTransform_mvnBuildFailed, but it only fires when maven exection fails.

I changed the name to make it more descriptive.

telemetry/definitions/commonDefinitions.json
mk-fan: That make sense. I removed the field.

telemetry/definitions/commonDefinitions.json
mk-fan: That make sense. I removed the field.

telemetry/definitions/commonDefinitions.json
mk-fan: As @dhasani23 suggested, this result is already captured by codeTransform_mvnBuildFailed, so I just removed this field.

telemetry/definitions/commonDefinitions.json
```diff
@@ -982,6 +982,11 @@
             "type": "string",
             "description": "The API request ID"
         },
+        {
```
damntrecky, line 985: Blocker: We can probably just re-use an existing field like `codeTransformRunTimeLatency` or a more generic one like `duration`. Right now all of our events use the `codeTransformRunTimeLatency`.

I would suggest against new types, unless necessary because we have to modify backend to receive new "types". But not "metrics".

damntrecky (approved):
Pre-approval, please address comments to remove new type field and use a more generic type field for logging duration in milliseconds.

telemetry/definitions/commonDefinitions.json
mk-fan: Changed to reuse "codeTransformRunTimeLatency".

telemetry/definitions/commonDefinitions.json
justinmk3: > The only related existing event is codeTransform_mvnBuildFailed, but it only fires when maven exection fails.

That's why that metric should be named `codeTransform_mvnBuild`. "Failed" or "success" are stored in the `result` field, it doesn't make sense to have a metric specifically for the "failed" case.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3808,6 +3808,20 @@
                 }
             ]
         },
+        {
+            "name": "codeTransform_mvnCommandsExecutionComplete",
```
justinmk3, line 3812: - What happens if you add `gradle` support? Are you going to create a bunch of `gradle` metrics? 
    - Why not store the "build tool" as a field in the metric instead?
- `ExecutionComplete` is another thing that just doesn't need to be spelled out in the metric name. Just call it `build`.

Based on the above, this metric could be named `codeTransform_build`, and it could have a `buildTool` field.

PR #704, opened 2024-02-28
feat: Update  amazonq_isAcceptedCodeChanges metric

dianaabv16:
## Problem

In amazonq we want to be able to capture number of rejected by the customer files, when accepting changes
## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -929,6 +929,11 @@
             "description": "Captures the number of files generated as a part of code generation iteration"
         },
         {
+            "name": "amazonqNumberOfFilesRejected",
+            "type": "double",
+            "description": "Captures the number of rejected files as a part of code generation iteration"
+        },
+        {
             "name": "amazonqRepositorySize",
             "type": "double",
             "description": "Captures the size of the source code"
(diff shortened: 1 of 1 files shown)


sannicm (approved):

jpinkney-aws (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -2864,6 +2869,7 @@
             "metadata": [
                 { "type": "enabled" }, 
                 { "type": "amazonqConversationId" },
+                { "type": "amazonqNumberOfFilesAccepted"},
```
justinmk3, line 2872: If the `amazonqNumberOfFilesGenerated` field were named `amazonqFiles`, it could have been re-used here. This is an example (one of many) of how over-specifying fields results in tons of very narrowly-useful fields, which causes lots of boilerplate and redundant things that everyone must read and figure out. It's not helpful, it's a maintenance cost for both data analysts + developers.

Is it too late to rename the `amazonqNumberOfFilesGenerated` field?

dianaabv16: I could be wrong, but I don't see how we would reused `amazonqFiles`:
`amazonqNumberOfFilesGenerated` -> is the number of file the customer got as the result of code generation
`amazonqNumberOfFilesAccepted`-> is the number of file the customer actually accepted out of generated files
Example: out of 5 generated files only 3 were accepted

justinmk3: They are both just numbers. The metric that they live on changes their "meaning". 

- `amazonq_isAcceptedCodeChanges.fileCount`
- `amazonq_codeGenerationInvoke`.fileCount`

There is no reason to repeat the entire metric name in each field name. The same generic field can be used in different metrics for different purposes.

PR #705, opened 2024-02-29
Validate that telemetry definitions are ordered alphabetically; Reorder the current file

awschristou:
## Problem

Metrics are defined in haphazard order in the telemetry definitions file. This makes it challenging to locate entries, particularly those sharing a common prefix. Maintaining sorted lists also makes it easier to reason about entries being added or modified, without having to think about their location.

## Solution

Scripting has been created to:
- test that the definition file is arranged in alphabetical order
- automatically rearrange the file's entries in alphabetical order

A CI job has been set up to run the validation check during PRs

This change was not wired up as a commit hook, because the repo does not have a top level node project. The CI job, and the readme are the mechanisms used to maintain the file going forward.

I chose not to use eslint. I started out in that direction, but it was becoming more of a distraction from the problem being solved. From https://github.com/azeemba/eslint-plugin-json?tab=readme-ov-file#is-eslint-really-the-best-tool-to-lint-my-json:
> Is **eslint** really the best tool to lint my JSON?
> Not really. `eslint` plugin interface wasn't designed to lint a completely different language but its interface is flexible enough to allow it. So this plugin is certainly unusual.

I ran the fix script against the current definitions file. This PR also contains that change as a separate commit.

To verify this change, I ran the new fix script on the current telemetry file. I then hand sorted the same file, and verified that the resulting file matched what the script produced.

## License


telemetry/README.md
```diff
@@ -74,7 +74,17 @@ and change `name`, `description`, and the values in `allowedValues`
...
+CI jobs run validation checks on the telemetry definitions file. For example, entries are expected to exist in alphabetical order. You may want to check this yourself before creating a PR, by switching to the `telemetry/validation` folder, and running:
+
+```
+npm install
+npm run validate
+```
+
+You can also run `npm run fix` to apply simple fixes to the file (like re-ordering).
+
 ### Overriding existing telemetry
```
dhruvigajjar, line 86: Thank you for updating docs

.github/workflows/lint-telemetry-definitions.yaml
```diff
@@ -0,0 +1,26 @@
+name: Lint Telemetry Definitions
```
justinmk3, line 1: could be just "Lint", all linting can be in one CI job
```suggestion
name: Lint
```

.github/workflows/lint-telemetry-definitions.yaml
```diff
@@ -0,0 +1,26 @@
...
+        name: Build and Test
+        runs-on: ubuntu-latest
+        steps:
+            - name: Sync Code
+              uses: actions/checkout@v3
+
+            - name: Compile the Validation code
+              run: |
+                  cd telemetry/validation
+                  npm install
```
justinmk3, line 19: seems like we should use the package-lock.json

```suggestion
                  npm ci
```

telemetry/validation/src/sorting.ts
```diff
@@ -0,0 +1,13 @@
+import { Metric, TelemetryField } from './telemetryDefinitions'
+
+export function fieldSort(a: TelemetryField, b: TelemetryField): number {
```
justinmk3, line 3: Why is the logic spread out across many files (reorder.ts, sorting.ts)? It could all in one file. The separation isn't really gaining anything, and forces new developers to build a mental-model of all the files

.github/workflows/lint-telemetry-definitions.yaml
awschristou: I'd prefer to parallelize all of the linting, having each job focus on its own domain. We only have the one thing being linted in CI, and it's easy to change course though. Let's leave it as-is for now.

telemetry/validation/src/fix.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const jsonPath = process.argv[2]
+    const definitions = await loadTelemetryDefinitions(jsonPath)
+
+    fix(definitions)
+
+    await saveTelemetryDefinitions(definitions, jsonPath)
+}
+
+export function fix(definitions: TelemetryDefinitions): void {
```
justinmk3, line 22: Seems like we're re-inventing eslint. We should probably implement this as an eslint plugin, otherwise it's a novel system that others must learn, for not much reason.

vscode toolkit has custom eslint plugins here: https://github.com/aws/aws-toolkit-vscode/tree/master/plugins/eslint-plugin-aws-toolkits

telemetry/validation/src/sorting.ts
awschristou: Naming is hard :)  I can consolidate them though, will take a shot at it

telemetry/validation/src/fix.ts
awschristou: Apparently its a little different to make eslint rules on json. I started out that way, and opted to just write pure scripts.

telemetry/validation/src/validation.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    for (let i = 0; i < definitions.metrics.length; i++) {
+        const inputMetric = definitions.metrics[i]
+        const sortedName = sortedNames[i]
+
+        if (inputMetric.name != sortedName) {
+            validations.push(`Telemetry Metrics are not sorted. Expected: ${sortedName}, Found: ${inputMetric.name}`)
+            break
+        }
+    }
```
dhruvigajjar, line 44: I might be being too picky here but should we also sort `types` used in `metrics`. e.g. codewishperer metrics have so many types.. 
https://github.com/aws/aws-toolkit-common/blob/c07e6a5a64381fcfaa89ef5130fe27451137ccfd/telemetry/definitions/commonDefinitions.json#L2969

awschristou: Yup, both `types` and `metrics` get sorted. There are some tests that exercise each of these.

telemetry/validation/src/validation.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    const sortedNames = definitions.metrics.map(t => t.name).sort(stringSort)
+
+    for (let i = 0; i < definitions.metrics.length; i++) {
+        const inputMetric = definitions.metrics[i]
+        const sortedName = sortedNames[i]
+
+        if (inputMetric.name != sortedName) {
+            validations.push(`Telemetry Metrics are not sorted. Expected: ${sortedName}, Found: ${inputMetric.name}`)
+            break
```
dhruvigajjar, line 42: should we post msg that they can fix by command? 
perhaps adding run `npm run fix`

dhruvigajjar: ohh my bad.. I see it's already added in final msg

telemetry/validation/src/validation.ts
dhruvigajjar: I might have missed the logic but i meant `types` defined in `metadata` of `metrics`

awschristou: Oh I see. Right, I only sorted the root `types` field. 

Yes, that is worthwhile for the same reasons. Let me take a look into that.

awschristou: Done

justinmk3 (approved):
This still seems a bit over-engineered. A validation script shouldn't need a package.json, it could be plain javascript. This will add a dependency-maintenance cost too, besides the more general risk that it's another project that will accumulate quirks and hidden corners. 

eslint plugins do require a sub-project. But with this solution we're gaining a project but not gaining eslint :/ (Assuming that eslint is even needed; if we are only "linting" json we don't need eslint, or a test framework...)

PR #706, opened 2024-03-06
telemetry: add new field for prev status

dhasani23:
## Problem

We already have a status field, but also want to be able to track the previous status of a job so that we can see how the status changes over time more easily. Using a new field rather than putting it all in the existing status field because our dashboards and other metrics are already configured to visualize the status with the existing field, so don't want to break that.

## Solution

Add new field to hold job's previous status.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -283,6 +283,11 @@
             ]
         },
         {
+            "name": "codeTransformPreviousStatus",
+            "type": "string",
+            "description": "The current transformation job's previous status"
+        },
+        {
             "name": "codeTransformRequestId",
             "type": "string",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -282,6 +282,11 @@
                 "RemoteRunProject"
             ]
         },
+        {
```
damntrecky, line 285: Question: Are you adding this new type to a metric? Will need to define that here as well.

dhasani23: Oh shoot forgot to do that, fixing now

mk-fan (approved):

dhruvigajjar (approved):

PR #716, opened 2024-03-28
auth: add isReAuth attribute to aws_loginWithBrowser

nkomonen-amazon:
This attribute indicates if the `aws_loginWithBrowser` process was done for reauthentication, otherwise regular authentication was done if isReauth !== true.

This gives us some extra details so we can separate user triggered authentications vs. system triggered reauthentications

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1059,6 +1059,11 @@
             "description": "Whether this was an individual point or an aggregation of points."
         },
         {
+            "name": "isReAuth",
+            "type": "boolean",
+            "description": "If this was performed as part of the reauthentication flow."
+        },
+        {
             "name": "lambdaArchitecture",
             "allowedValues": [
                 "x86_64",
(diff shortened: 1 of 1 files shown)


justinmk3:
> separate user triggered authentications vs. system triggered reauthentications

Wouldn't `passive` indicate that ? Or at least, it should, if it doesn't already.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1058,6 +1058,11 @@
             "type": "boolean",
             "description": "Whether this was an individual point or an aggregation of points."
         },
+        {
+            "name": "isReAuth",
+            "type": "boolean",
+            "description": "If reauthentication was done"
```
justinmk3, line 1064: What does this mean?
```suggestion
            "description": "If an existing connection was reauthenticated (as opposed to ...)"
```

nkomonen-amazon: Couldn't really think of a good description. This is basically a flag for if reauthentication was done during the createToken process. I've reworded, lmk if it improves.

justinmk3: Main goal is to help other Toolkits apply this, so any description that helps them understand when this is relevant, is helpful. 

Mentioning "during createToken" is probably worth adding

justinmk3 (approved):

PR #718, opened 2024-04-04
telemetry(amazonq): add field to track auth type to metric

dhasani23:
## Problem

Want a field to track if user is authenticated with BuilderId or IdC for QCT.

## Solution

Use an existing auth-related field and add it to an existing metric which will be emitted when /transform is invoked.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2841,6 +2841,10 @@
                 {
                     "type": "codeTransformSessionId",
                     "required": true
+                },
+                {
+                    "type": "enabledAuthConnections",
+                    "required": false
                 }
             ]
         },
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2843,7 +2843,7 @@
                     "required": true
                 },
                 {
-                    "type": "enabledAuthConnections",
+                    "type": "credentialSourceId",
                     "required": false
                 }
             ]


telemetry/definitions/commonDefinitions.json
```diff
@@ -2841,6 +2841,10 @@
                 {
                     "type": "codeTransformSessionId",
                     "required": true
+                },
```
damntrecky, line 2844: Here is the definition for that type:

```
        {
            "name": "enabledAuthConnections",
            "type": "string",
            "description": "Comma delimited list of enabled auth connections"
        },
        ```

damntrecky (approved):
Tentative approval: moving discussion into slack

JuliaABurch (approved):

telemetry/definitions/commonDefinitions.json
dhasani23: Discussed in Slack: going to use `jobIsStartedFromChatPrompt` for now since `projectDetails` is always emitted on IDE startup; we want this metric to capture just users who tried transforming a valid project.

justinmk3:
> Want a field to track if user is authenticated with BuilderId or IdC for QCT.

`credentialSourceId` can be used for that (`awsId` = "builder id"): https://github.com/aws/aws-toolkit-common/blob/b7f05546af2d084ebf36b98c837ddf3641e3a63f/telemetry/definitions/commonDefinitions.json#L763-L776

PR #729, opened 2024-05-03
telemtry:(amazonq): Add more metadata fields to capture duration, result, failure reason to amazonq events

sarayev:
## Problem

We want to track duration, result, failure reason, requestId and requestServiceType fields for startConversationInvoke, approachInvoke, createUpload and codeGenerationInvoke events in JetBrains. 

## Solution

In toolkits, we get those fields added to the telemetry events when we do telemetry.event.run{}, however in JetBrains it isn't supported. Therefore, adding them as metadata fields.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1340,6 +1340,26 @@
                 {
                     "type": "credentialStartUrl",
                     "required": false
+                },
+                {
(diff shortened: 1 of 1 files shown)


jsalles (approved):

justinmk3 (approved):
Thank you for using *existing* fields instead of creating special-purpose fields that are semantically the same as existing fields.

PR #731, opened 2024-05-06
telemetry(amazonq): source folder modification

sannicm:
## Problem

The source folder modification was not being tracked, it was evident when a bug was spotted and it was not possible to measure any customer impact. This should make it easier when integrated with the toolkits to measure usage and impact if something goes wrong.

## Solution

Add modifySourceFolder metric.

It was tested in the VSCode toolkit adding the metric in vscodeTelemetry.json file, generating telemetry and then checking it was emitted in kibana.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1599,6 +1599,19 @@
             ]
         },
         {
+            "name": "amazonq_modifySourceFolder",
+            "description": "User modified source folder",
+            "metadata": [
+                {
+                    "type": "amazonqConversationId"
+                },
+                {
+                    "type": "credentialStartUrl",
+                    "required": false
+                }
+            ]
+        },
+        {
             "name": "amazonq_startConversationInvoke",
             "description": "Captures startConversation invocation process",
             "metadata": [


telemetry/definitions/commonDefinitions.json
```diff
@@ -1598,6 +1598,20 @@
                 }
             ]
         },
+        {
+            "name": "amazonq_modifySourceFolder",
+            "description": "User modified the source folder as the current context was too large",
+            "unit": "Count",
```
jsalles, line 1604: What does `"unit": "Count"` mean? How's it different from not specifying "unit" at all?

justinmk3: unit decides the form of the standard `value` field which is present on all metrics by default https://github.com/aws/aws-toolkit-common/blob/78f6b61320627e7bfc60c21cd27991b4eb86db63/telemetry/telemetryformat.md#global-arguments

telemetry/definitions/commonDefinitions.json
```diff
@@ -1598,6 +1598,20 @@
                 }
             ]
         },
+        {
+            "name": "amazonq_modifySourceFolder",
+            "description": "User modified the source folder as the current context was too large",
```
justinmk3, line 1603: "as the current context was too large" sounds like over-specification (or I just don't understand what it means, which is a different problem). Does the metric need to be so narrowly defined?

justinmk3:
what does "The source folder modification" mean? Is this referring to when the user changes the vscode workspace folder, or something else? 

If it's referring to a vscode operation, then the metric should have a more generic prefix such as `ide_` or `vscode_`

telemetry/definitions/commonDefinitions.json
sannicm: That's a fair point, It does not, it will make sense for our team at least. I'll reduce it

telemetry/definitions/commonDefinitions.json
sannicm: @jsalles in addition, I'm not using the "enabled" field which I saw as not needed but is in several other events. In this case, just counting the usage + clientId will give us information on impact or even on usage.

justinmk3:
> It referst to a /dev feature to let the customer reduce the amount of folders (reducing the context) to upload for code generation. The usage was not being tracked and there was a bug on this feature

Wouldn't that make sense as some sort of "count" field on the existing on the "upload" or "generate" metric (`amazonq_codeGenerationInvoke` or whatever the right one is)? Why emit a new metric for this?

sannicm:
> > It referst to a /dev feature to let the customer reduce the amount of folders (reducing the context) to upload for code generation. The usage was not being tracked and there was a bug on this feature
> 
> Wouldn't that make sense as some sort of "count" field on the existing on the "upload" or "generate" metric (`amazonq_codeGenerationInvoke` or whatever the right one is)? Why emit a new metric for this?

@justinmk3 This is completely unrelated to the code generation invocation or planning. /dev is calculating the size of the repo before uploading and there is no visibility on the usage of the feature that allows the customer to modify a source folder for retrying generation. This will emit a metric when specficially this feature is used.

telemetry/definitions/commonDefinitions.json
sannicm: @justinmk3 the` "unit": "Count"` is not needed. I misunderstood how the visualization in kibana was working. Now the metric should be correct.

jsalles (approved):

justinmk3:
> > Wouldn't that make sense as some sort of "count" field on the existing on the "upload" or "generate" metric (`amazonq_codeGenerationInvoke` or whatever the right one is)? Why emit a new metric for this?
> 
> This is completely unrelated to the code generation invocation or planning. /dev is calculating the size of the repo before uploading and there is no visibility on the usage of the feature that allows the customer to modify a source folder for retrying generation. This will emit a metric when specficially this feature is used.

The user action in that case, semantically, is `/dev/`. The user initiated that, right? Isn't there a metric already associated with that? (`amazonq_codeGenerationInvoke` was just a guess; I'm hoping you will know the actually relevant metric that already captures the general case when a user invokes `/dev/`)

telemetry/definitions/commonDefinitions.json
```diff
@@ -1598,6 +1598,19 @@
...
+            "name": "amazonq_modifySourceFolder",
+            "description": "User modified source folder",
+            "metadata": [
+                {
+                    "type": "amazonqConversationId"
+                },
+                {
+                    "type": "credentialStartUrl",
+                    "required": false
+                }
```
justinmk3, line 1611: Since these are the only fields, this indicates that this metric is literally just a boolean. Why can't it be a field on an existing metric?

telemetry/definitions/commonDefinitions.json
```diff
@@ -1598,6 +1598,19 @@
                 }
             ]
         },
+        {
+            "name": "amazonq_modifySourceFolder",
+            "description": "User modified source folder",
```
justinmk3, line 1603: > This could happen on different steps of our process, whenever we upload the repo to our backend (that happens at least once before planning the execution, and once more for each code generation requested).
> It is kinda like a sub-command of `/dev` but it’s not triggered by the user whenever they want, and only when we think the repo is too large and a sub-folder should be selected.
> We display the button and want to capture the action of the user clicking it.

PR #734, opened 2024-05-09
telemetry(amazonq): add field for ZIP size and remove unused metrics

dhasani23:
## Problem

Need to track the size of the upload ZIP even when the upload itself fails so we have some more info. Also had some metrics which were unused in both IDEs.

## Solution

Add `codeTransformTotalByteSize` field to existing metric and remove unused metrics.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2762,16 +2762,6 @@
             "passive": true
         },
         {
-            "name": "codeTransform_configurationFileSelectedChanged",
-            "description": "Fire a change event when users select a new configuration/module file from the dropdown.",
-            "metadata": [
-                {
-                    "type": "codeTransformSessionId",
-                    "required": true
-                }
-            ]
-        },
-        {
             "name": "codeTransform_dependenciesCopied",
             "description": "The repo copies over at least one dependency successfully.",
             "metadata": [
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -2825,20 +2815,6 @@
                 }
             ]
         },
-        {
-            "name": "codeTransform_isDoubleClickedToTriggerUserModal",
-            "description": "The code transform button in the devtools to initiate the pre transform user modal window.",
-            "metadata": [
-                {
-                    "type": "codeTransformSessionId",
-                    "required": true
```
dhruvigajjar, line 2834: by removing required:true types, are these metrics already deprecated in implementation?

dhasani23: Yes none of these metrics are used right now.

damntrecky: Double check: Does that exist for both Jetbrains and vscode? This will be a breaking change otherwise.

dhasani23: Double checked and none of these metrics are currently used in JetBrains nor VS Code. As long as HIL didn't re-introduce any of these this change should be fine to make.

justinmk3:
Thanks for removing the unused metrics, that is helpful.

PR #735, opened 2024-05-09
telemetry(IamPolicyChecks): create policy checks metrics

kevluu-aws:
Add three new metrics for new IAM Policy Checks feature in AWS Toolkit VS Code.
One for each execution type: validate policy and custom policy checks
One for emitting error metrics

This feature uses a webview as its main interface, which will use UiClick on applicable elements.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -133,6 +133,20 @@
             "description": "High level categorization indicating the cause of the error"
         },
         {
+            "name": "cfnParameterFileUsed",
+            "type": "boolean",
+            "description": "Boolean value of whether or not a Cfn parameter file is provided."
+        },
+        {
+            "name": "checkType",
+            "type": "string",
+            "allowedValues": [
+                "CheckNoNewAccess",
+                "CheckAccessNotGranted"
+            ],
+            "description": "User inputted check type to denote which custom check to run."
+        },
+        {
             "name": "cloudWatchLogsPresentation",
             "allowedValues": [
                 "ui",
@@ -838,6 +852,16 @@
             "description": "The id of the detector which produced the code scan issue"
         },
         {
+            "name": "documentType",
+            "type": "string",
+            "allowedValues": [
+                "CloudFormation",
+                "Terraform Plan",
+                "JSON Policy Language"
+            ],
+            "description": "Document type of the edited file in IAM Policy Checks"
+        },
+        {
             "name": "duration",
             "type": "double",
             "description": "The duration of the operation in milliseconds"
@@ -990,6 +1014,11 @@
             "description": "The id of a security finding from a code scan"
         },
         {
+            "name": "findingsCount",
+            "type": "int",
+            "description": "Number of findings discovered after executing IAM Policy Checks"
+        },
+        {
             "name": "framework",
             "type": "string",
             "description": "Application framework being used"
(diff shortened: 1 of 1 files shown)


mponaws (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -1324,6 +1373,63 @@
...
+                {
+                    "type": "result"
+                }
+            ]
+        },
+        {
+            "name": "accessanalyzer_iamPolicyChecksError",
+            "description": "Errors related to IamPolicyChecks",
+            "metadata": [
+                {
```
dhruvigajjar, line 1407: why can we combine for Error with validation? 
result field can tell us if validation was successful for not. Also Error can have same metadata info as `accessanalyzer_iamPolicyChecksValidatePolicy` as reason, source, policy size, fileUsed, etc..

kevluu-aws: We have some errors that are not always related to ValidatePolicy or CustomChecks. This would allow us to congregate all errors across the board, rather than looking at ValidatePolicy Failed, CustomChecks Failed, etc.

dhruvigajjar: Idea is to cover both cases of metric in one. Meaning `ValidatePolicy` will have both success & failed result, `CustomChecks` should cover success & failed result. If you feel the need to explicitly have Error metric, then we can discuss.

kevluu-aws: Removed Error metric, using ValidatePolicy and CustomChecks to cover both results.

telemetry/definitions/commonDefinitions.json
```diff
@@ -837,6 +851,16 @@
             "type": "string",
             "description": "The id of the detector which produced the code scan issue"
         },
+        {
+            "name": "documentType",
+            "type": "string",
+            "allowedValues": [
+                "CloudFormation",
+                "Terraform Plan",
```
dhruvigajjar, line 859: will there be only these document types? do you want to include "Other"

telemetry/definitions/commonDefinitions.json
```diff
@@ -989,6 +1013,11 @@
             "type": "string",
             "description": "The id of a security finding from a code scan"
         },
+        {
+            "name": "findingsCount",
+            "type": "int",
+            "description": "Number of findings discovered after execution"
```
dhruvigajjar, line 1019: can you be more specific about what kind of execution?

telemetry/definitions/commonDefinitions.json
```diff
@@ -1038,6 +1067,16 @@
             "type": "boolean",
             "description": "Whether or not the deploy targets a new destination (true) or an existing destination (false)"
         },
+        {
+            "name": "inputPolicyType",
+            "type": "string",
+            "allowedValues": [
+                "Identity",
```
dhruvigajjar, line 1074: same as above comment. do we want to include "Other"

dhruvigajjar: Maybe you can combine policy type and in metric say if it was input or reference. I might not have full context of use-case so feel free to discard

telemetry/definitions/commonDefinitions.json
dhruvigajjar: looks good

telemetry/definitions/commonDefinitions.json
kevluu-aws: They refer to different things in the feature, my opinion is to keep it separate

kevluu-aws: As for this we will only have Identity or Resource, so "Other" feels not so useful

telemetry/definitions/commonDefinitions.json
kevluu-aws: Updated it to "Number of findings discovered after executing IAM Policy Checks"

telemetry/definitions/commonDefinitions.json
dhruvigajjar: Got it

dhruvigajjar (approved):

PR #736, opened 2024-05-16
adding scopes to auth metrics for detailed authStatus

dhruvigajjar:
## Problem

Adding auth telemetry for to identify user's auth Status upon start up.

## Solution
auth_userState list scopes upon users loading extension so we know their authStatus upon startup

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -103,6 +103,22 @@
             "description": "Current number of authentication connections the user has"
         },
         {
+            "name": "authEnabledConnections",
+            "type": "string",
+            "description": "Comma-delimited list of enabled auths."
+        },
+        {
+            "name": "authScopes",
+            "type": "string",
+            "description": "Comma-delimited list of scopes that users has"
+        },
+        {
+            "name": "authStatus",
+            "type": "string",
+            "allowedValues": ["connected", "notConnected", "expired"],
+            "description": "Status of the an auth connection."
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1848,6 +1872,29 @@
                 }
             ]
         },
+        {
+            "name": "auth_userState",
+            "description": "The state of the user's authentication.",
```
justinmk3, line 1877: metrics are for events. state isn't an event

What event does auth_userState represent? It looks more like data. Can the data be stored on an existing event? Otherwise, it's not clear when this "event" should be emitted, it would be more like a heartbeat (which then implies it should be on... an existing, semi-frequent event)

dhruvigajjar: it's only on startup.

justinmk3: I see that this is upstreaming an existing metric: https://github.com/aws/aws-toolkit-vscode/blob/0f9782a9d8407d3d755c44dc3a3eec1fe405e7b9/packages/core/src/shared/telemetry/vscodeTelemetry.json#L1108-L1109

dhruvigajjar: yes Ideally it should be in common only because JetBrains is also emitting this metric. Trying to fix it by moving it to common.

justinmk3 (approved):

PR #754, opened 2024-06-17
Add requestId and reasonDesc to refreshCredentials metric

wularr:
## Problem
refreshCredentials metric does not have requestId and reason description

## Solution

Add requestId and reasonDesc to refreshCredentials 

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2202,6 +2202,14 @@
                     "type": "result"
                 },
                 {
+                    "type": "requestId",
+                    "required": false
+                },
+                {
+                    "type": "reasonDesc",
+                    "required": false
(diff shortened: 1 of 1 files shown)


justinmk3:
"reasonDesc" needs to be defined. you can add it next to "reason": https://github.com/aws/aws-toolkit-common/blob/a88c20cf3dfef5971614327700d2b117855a3eb6/telemetry/definitions/commonDefinitions.json#L1196-L1200

something like this:

```
        {
            "name": "reasonDesc",
            "type": "string",
            "description": "Error message detail (truncated). May contain arbitrary message details (as opposed to the 'reason' field, which is intended to have a predictable value for a particular class of failures)."
        },
```

dhruvigajjar (approved):

PR #767, opened 2024-06-27
auth: Add reAuthReason field in aws_loginWithBrowser

nkomonen-amazon:
When the user logs in with the browser AND it is due to reauth, we want to know why they need to reauth. This will explain why.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1286,6 +1286,11 @@
             "description": "Error message detail. May contain arbitrary message details (unlike the `reason` field), but should be truncated (recommendation: 200 chars)."
         },
         {
+            "name": "reAuthReason",
+            "type": "string",
+            "description": "Why was the connection/session put in to a state where it needs to be reauthenticated?"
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1285,6 +1285,11 @@
             "type": "string",
             "description": "Error message detail. May contain arbitrary message details (unlike the `reason` field), but should be truncated (recommendation: 200 chars)."
         },
+        {
+            "name": "reAuthReason",
```
jpinkney-aws, line 1289: is this something that could eventually be pulled out into common?

jpinkney-aws (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -2283,6 +2288,10 @@
                     "type": "isReAuth",
                     "required": false
                 },
+                {
+                    "type": "reAuthReason",
```
justinmk3, line 2292: why not re-use the `reason` field?

PR #770, opened 2024-06-28
Add fields for codeTransform_jobStart metric

sunwoo-ms:
Add metric types for Code Transform to be fired when a transformation is started from the Solution Explorer and when a build is completed. Modified existing metric (`codeTransform_jobIsStartedFromChatPrompt` -> `codeTransform_jobStart`) to also record the number of times a transformation plan was received during the transformation process.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -333,6 +333,11 @@
             ]
         },
         {
+            "name": "codeTransformPlanReceivedCount",
+            "type": "int",
+            "description": "The number of times the transformation plan was received during transformation"
+        },
+        {
             "name": "codeTransformPreValidationError",
             "type": "string",
             "description": "Names of the pre-validation errors that can occur",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -3313,6 +3313,20 @@
             ]
         },
         {
+            "name": "codeTransform_jobIsStartedFromSolutionExplorer",
+            "description": "The user initiates a transform job from the Solution Explorer.",
+            "metadata": [
+                {
+                    "type": "codeTransformSessionId",
+                    "required": true
+                },
+                {
+                    "type": "credentialSourceId",
+                    "required": false
+                }
+            ]
+        },
+        {
             "name": "codeTransform_jobStartedCompleteFromPopupDialog",
             "passive": true,
             "description": "After modal validation of inputs, code execution will start and we will fire this event.",
@@ -3333,6 +3347,21 @@
                     "type": "codeTransformSessionId",
                     "required": true
                 }
+            ]
+        },
+        {
+            "name": "codeTransform_buildCompleted",
+            "passive": true,
+            "description": "During the transformation progress, log whether the build succeeded or failed.",
+            "metadata": [
+                {
+                    "type": "buildStatus",
+                    "required": true
+                },
+                {
+                    "type": "codeTransformSessionId",
+                    "required": true
+                },
             ]
         },
         {
@@ -3503,6 +3532,10 @@
                 },
                 {
                     "type": "codeTransformLocalMavenVersion",
+                    "required": false
+                },
+                {
+                    "type": "codeTransformPlanReceivedCount",
                     "required": false
                 },
                 {
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -389,11 +389,6 @@
             "description": "Represents the IDE session from which users start the transformation process"
         },
         {
-            "name": "codeTransformSolutionSelected",
-            "type": "boolean",
-            "description": "Indicates whether the solution is the selected transformation target (true if yes, false if no)"
-        },
-        {
             "name": "codeTransformStartSrcComponents",
             "type": "string",
             "description": "Names of components that can start a transformation",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -3335,6 +3349,21 @@
                 }
             ]
         },
+        {
+            "name": "codeTransform_buildCompleted",
```
justinmk3, line 3353: same comment as above.

there is an existing `codeTransform_jobStatusChanged` event. Can "build completed" be a kind of status stored on that existing event?

I see also that gumby defined a `codeTransform_mvnBuildFailed` event. This, again, doesn't make sense. 

Instead, I suggest:

- a `codeTransform_build` event
- it has a `tool` field with values like "maven", "gradle" etc.
- its `result` field can store "success" or "failed. this is a standard field on every metric.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3312,6 +3312,20 @@
                 }
             ]
         },
+        {
+            "name": "codeTransform_jobIsStartedFromSolutionExplorer",
```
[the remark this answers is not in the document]
sunwoo-ms: Thank you for your feedback! I've addressed this in a more recent commit.

telemetry/definitions/commonDefinitions.json
sunwoo-ms: Thank you for your feedback! I've addressed this in a more recent commit.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3309,6 +3309,14 @@
                 {
                     "type": "credentialSourceId",
                     "required": false
+                },
+                {
+                    "type": "isStartedFrom",
```
awschristou, line 3314: If you want a specific enum, you'll have to define it in types. Here, you can probably use the type `source`, which is an open ended string. Then in the Toolkit, we use specific constant values. This keeps the metadata field open ended to support future use cases as well.

I see you have source immediately below this. You might ne able to remove this field.
How would source and isStartedFrom differ?

telemetry/definitions/commonDefinitions.json
```diff
@@ -3335,6 +3343,17 @@
                 }
             ]
         },
+        {
+            "name": "codeTransform_build",
+            "passive": true,
+            "description": "During the transformation progress, log the status of the build.",
+            "metadata": [
+                {
+                    "type": "tool",
```
awschristou, line 3352: What does this represent - Could this also be replaced with "source"?

telemetry/definitions/commonDefinitions.json
```diff
@@ -3505,6 +3524,10 @@
                     "type": "codeTransformLocalMavenVersion",
                     "required": false
                 },
+                {
+                    "type": "codeTransformPlanReceivedCount",
```
awschristou, line 3528: You have to add all distinct types into the types section at the top of the file.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3299,8 +3299,8 @@
             ]
         },
         {
-            "name": "codeTransform_jobIsStartedFromChatPrompt",
-            "description": "The user initiates a transform job from Amazon Q chat prompt.",
+            "name": "codeTransform_jobIsStarted",
```
justinmk3, line 3302:
```suggestion
            "name": "codeTransform_jobStart",
```

telemetry/definitions/commonDefinitions.json
sunwoo-ms: This has been addressed in a more recent commit.

telemetry/definitions/commonDefinitions.json
sunwoo-ms: You're correct, this is a redundant field. We've removed it in a more recent commit.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3299,16 +3335,32 @@
...
+                    "type": "codeTransformNumberOfProjects",
+                    "required": true
+                },
                 {
                     "type": "codeTransformSessionId",
                     "required": true
                 },
+                {
+                    "type": "codeTransformSolutionSelected",
+                    "required": true
```
awschristou, line 3351: Since this is a generalized metric to be used from different Toolkits (IDEs), you have to consider if certain properties will be relevant to each product. Transforming Java in IntelliJ may not require this. Can you check the required field for all of the properties you added or changed? It's okay if you want VS to always use this field, there just won't be any programmatic enforcement.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3299,16 +3335,32 @@
...
+                {
+                    "type": "codeTransformTarget",
+                    "required": true
+                },
                 {
                     "type": "credentialSourceId",
                     "required": false
+                },
+                {
+                    "type": "source",
```
awschristou, line 3362: Would it be possible to represent codeTransformSolutionSelected through `source`? I wonder if you can avoid adding a new property/type

sunwoo-ms: What would be an example of `source` as you think it should be used?

telemetry/definitions/commonDefinitions.json
sunwoo-ms: I see what you're getting at. A different IDE wouldn't necessarily be transforming projects and solutions. I can make it so that all fields relating to projects and solutions are optional.

telemetry/definitions/commonDefinitions.json
sunwoo-ms: This has been addressed in a more recent commit.

telemetry/definitions/commonDefinitions.json
sunwoo-ms: This has been addressed in a more recent commit.

awschristou (approved):

PR #783, opened 2024-07-11
telemetry(amazonq): add/change metrics for gradle support

dhasani23:
## Problem

We're adding Gradle support for /transform and need to make some telemetry changes to reflect this.

## Solution

Make said changes. Also, simplify our telemetry by combining duplicate metrics and removing unneeded ones.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -3167,19 +3167,27 @@
             "passive": true
         },
         {
-            "name": "codeTransform_buildComplete",
+            "name": "codeTransform_buildResult",
             "passive": true,
-            "description": "During the transformation progress, log the status of the build.",
+            "description": "The result of the local build.",
             "metadata": [
                 {
-                    "type": "buildDuration",
+                    "type": "codeTransformBuildCommand",
                     "required": true
+                },
+                {
+                    "type": "buildDuration",
+                    "required": false
+                },
+                {
+                    "type": "codeTransformSessionId",
+                    "required": false
                 }
             ]
         },
         {
             "name": "codeTransform_dependenciesCopied",
-            "description": "The repo copies over at least one dependency successfully.",
+            "description": "At least one dependency is copied over successfully.",
             "metadata": [
                 {
                     "type": "codeTransformSessionId",
@@ -3464,24 +3472,6 @@
                 {
                     "type": "codeTransformSessionId",
                     "required": true
-                }
-            ]
-        },
-        {
-            "name": "codeTransform_buildFailed",
-            "description": "The local build failed.",
-            "metadata": [
-                {
-                    "type": "codeTransformBuildCommand",
-                    "required": true
-                },
-                {
-                    "type": "codeTransformSessionId",
-                    "required": true
-                },
-                {
-                    "type": "reason",
-                    "required": false
                 }
             ]
         },
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -244,9 +244,9 @@
             "description": "Any API-specific errors"
         },
         {
-            "name": "codeTransformApiNames",
+            "name": "codeTransformApiName",
             "type": "string",
-            "description": "Names of allowed API calls",
+            "description": "Name of the API",
             "allowedValues": [
                 "ResumeTransformation",
                 "StartTransformation",
@@ -430,16 +430,6 @@
             "name": "codeTransformUploadId",
             "type": "string",
             "description": "A unique upload ID for S3"
-        },
-        {
-            "name": "codeTransformVCSViewerSrcComponents",
-            "type": "string",
-            "description": "Names of components that can initiate the diff viewer",
-            "allowedValues": [
-                "toastNotification",
-                "treeView",
-                "treeViewHeader"
-            ]
         },
         {
             "name": "codewhispererAcceptedTokens",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -298,19 +298,32 @@
...
                 "mvnw.cmd",
                 "mvnw",
                 "mvn",
-                "IDEBundledMaven"
+                "IDEBundledMaven",
+                "gradlew.bat",
+                "gradlew",
+                "gradle",
+                "IDEBundledGradle"
+            ]
```
tincheng, line 318: For telemetry, is it necessary to enforce a list of allowed values? I think not imposing allowed values will make the metric types more scalable. Same for codeTransformBuildSystem

justinmk3: These enum-style metric definitions are purely for developer (client-side) convenience, they aren't enforced by the telemetry service. You can always override them in the client.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3488,11 +3505,11 @@
             ]
         },
         {
-            "name": "codeTransform_mvnBuildFailed",
-            "description": "The maven command build failed.",
+            "name": "codeTransform_buildFailed",
```
justinmk3, line 3508: Generalizing this as you've done is a great idea. Any reason not to go a bit further and generalize it to `codeTransform_build` ? Then the "failed" state can simply be represented by the `result` field.

Using standard fields allows dashboards to better indicate the health of the clients. (Also it's a bit strange that `codeTransform_buildFailed` would have a `result=Success` value).

dhasani23: Good point; will do that.

telemetry/definitions/commonDefinitions.json
```diff
@@ -418,16 +431,6 @@
             "type": "string",
             "description": "A unique upload ID for S3"
         },
-        {
-            "name": "codeTransformVCSViewerSrcComponents",
```
dhasani23, line 422: Not sure why we have this, it's not particularly useful and N/A at least for VS Code, where the diff viewer is only accessed from the single Download button.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3403,17 +3418,17 @@
             ]
         },
         {
-            "name": "codeTransform_logApiError",
+            "name": "codeTransform_logApiResult",
```
dhasani23, line 3421: Combines `logApiError` and `logApiLatency`.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3581,32 +3551,6 @@
...
-                    "required": true
-                },
-                {
-                    "type": "codeTransformStatus",
-                    "required": true
-                }
-            ]
-        },
         {
             "name": "codeTransform_vcsViewerClicked",
```
dhasani23, line 3555: Get rid of `vcsViewerCanceled` and `vcsViewerSubmitted`; repurpose `vcsViewerClicked` to fill in for these 2 removals.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3487,24 +3475,6 @@
...
-                {
-                    "type": "codeTransformMavenBuildCommand",
-                    "required": true
-                },
-                {
-                    "type": "codeTransformSessionId",
-                    "required": true
-                },
-                {
-                    "type": "reason",
```
dhasani23, line 3503: As a general question, we don't need to manually put `result` & `reason` in our metadata fields, right?

dhasani23:
A future follow-up to this PR is to take all of the fields and put them all in `codeTransformMetadata`

telemetry/definitions/commonDefinitions.json
tincheng: Should we leave logApiError/logApiLatency as is for backward compatibility, then deprecate them once no more metrics using these names are being published?

dhasani23: That makes sense

telemetry/definitions/commonDefinitions.json
tincheng: I think we are missing "chat" in the allowedValues. This metric can help us identity if most people uses chat to trigger diff viewer, or through transformation hub, or through notification pop-up. (Might have been more relevant in IntelliJ)

telemetry/definitions/commonDefinitions.json
```diff
@@ -293,36 +317,20 @@
...
-            ]
+            "description": "The Java version on the user's machine"
         },
         {
             "name": "codeTransformMetadata",
             "type": "string",
             "description": "A general field for logging metadata associated to Amazon Q transform metrics."
         },
-        {
-            "name": "codeTransformNumberOfProjects",
```
dhasani23, line 322: This and the few below were all unused fields.

telemetry/definitions/commonDefinitions.json
justinmk3: correct. At least not for vscode/visualstudio. I'm not sure about jetbrains, but it would be a pretty simple update to the generator , if needed. cc @manodnyab @rli

telemetry/definitions/commonDefinitions.json
```diff
@@ -3403,43 +3381,16 @@
...
-                {
-                    "type": "codeTransformTotalByteSize",
-                    "required": false
-                }
-            ]
-        },
-        {
-            "name": "codeTransform_logApiLatency",
-            "passive": true,
-            "description": "A general logging measure for all client side API's latency.",
```
justinmk3, line 3439: If you decide to keep these around, may want to change the description to:

    "description": "Deprecated. Use codeTransform_logApiResult instead.",

telemetry/definitions/commonDefinitions.json
```diff
@@ -3334,33 +3330,15 @@
                 }
             ]
         },
-        {
```
tincheng, line 3337: This metrics is being used by Visual Studio team

justinmk3: These metrics define the generated client(s); removing metrics here doesn't prevent old clients (e.g. old versions of the VS extension), from sending the old metrics. 

The next time VS updates its client, it just needs to migrate to the new metric.

dhasani23:
Closed in favor of: https://github.com/aws/aws-toolkit-common/pull/790

PR #784, opened 2024-07-11
telemetry: add Q dev bundle file metric definition

willyyhuang:
## Problem

We (feature dev team) want to understand if there are any other relevant code file extension that should be added to the allowlist.

## Solution

To obtain such information, we want to emit telemetry event to see the most common file extensions being ignored while bundling.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -835,6 +835,11 @@
             "description": "Time taken for context truncation in milliseconds"
         },
         {
+            "name": "count",
+            "type": "int",
+            "description": "Number of occurrences a metric, or some other metric-defined count."
+        },
+        {
             "name": "credentialModification",
             "type": "string",
             "allowedValues": [
@@ -1630,6 +1635,18 @@
                 {
                     "type": "credentialStartUrl",
                     "required": false
+                }
+            ]
+        },
+        {
+            "name": "amazonq_bundleExtensionIgnored",
+            "description": "File extension was ignored 'count' many times during bundling process",
+            "metadata": [
+                {
+                    "type": "count"
+                },
+                {
+                    "type": "filenameExt"
                 }
             ]
         },
@@ -6006,7 +6023,7 @@
                     "type": "attempts",
                     "required": false
                 },
-                { 
+                {
                     "type": "duration",
                     "required": false
                 },


chengoramazon (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -15,6 +15,16 @@
             "type": "double",
             "description": "Total time from start to end"
         },
+        {
+            "name": "amazonqFileExtensionName",
```
justinmk3, line 19: is there a reason the existing filenameExt field can't be used? https://github.com/aws/aws-toolkit-common/blob/7b972e3fe7ee01b3b3d99c4198915c1520c0401e/telemetry/definitions/commonDefinitions.json#L1129

please check for existing fields instead of adding new ones. same for the `amazonQFileIgnoredCount` field.

willyyhuang: no reason, let me adjust, will keep the `amazonQFileIgnoredCount` as it has a different use case/meaning

telemetry/definitions/commonDefinitions.json
```diff
@@ -15,6 +15,11 @@
             "type": "double",
             "description": "Total time from start to end"
         },
+        {
+            "name": "amazonQFileIgnoredCount",
```
justinmk3, line 19:
```suggestion
            "name": "count",
```

this will avoid needing to add a new field every time a metric wants to count something

telemetry/definitions/commonDefinitions.json
```diff
@@ -15,6 +15,11 @@
...
+            "type": "int",
+            "description": "Number of times that a file extension is ignored while bundling in feature dev."
```
justinmk3, line 21: this can be expressed by the metric itself. doesn't need to be built into the field.
```suggestion
            "description": "Number of occurrences a metric, or some other metric-defined count."
```

justinmk3, line 21: this can be expressed by the metric itself. doesn't need to be built into the field.

```suggestion
            "name": "count",
            "type": "int",
            "description": "Number of occurrences a metric, or some other metric-defined count."
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -1633,6 +1638,18 @@
                 }
             ]
         },
+        {
+            "name": "amazonq_bundleExtensionIgnored",
+            "description": "Captures file extensions ignored during bundling process",
```
justinmk3, line 1643:
```suggestion
            "description": "File extension was ignored 'count' many times during bundling process",
```

telemetry/definitions/commonDefinitions.json
willyyhuang: done!

PR #790, opened 2024-07-17
Define new metrics for Code Transform

tincheng:
## Problem
1. Existing Code Transform metrics do not follow naming convention or telemetry best practices.
2. Code Transform needs to support additional build tool (E.g. Gradle)

## Solution
Introduced new metrics for more generic usage, and to better track user journey through transformation. These metrics will replace some of the existing ones (mark as deprecated for removal later).

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -3430,23 +3488,42 @@
             "passive": true
         },
         {
-            "name": "codeTransform_buildComplete",
```
tincheng, line 3433: This metric was added for Visual Studio, but confirmed not being used, so we can clean it up.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3788,7 +3936,7 @@
         {
             "name": "codeTransform_projectDetails",
             "passive": true,
-            "description": "Transform initiates project validation before user interaction and we want to gather project details then.",
+            "description": "Transform initiates project validation before user interaction and we want to gather project details then. (Deprecated - use codeTransform_validateProject)",
```
justinmk3, line 3939: Let's make sure the developer cannot miss it:

```suggestion
            "description": "Deprecated - use codeTransform_validateProject",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -3769,7 +3917,7 @@
         },
         {
             "name": "codeTransform_mvnBuildFailed",
-            "description": "The maven command build failed.",
+            "description": "The maven command build failed. (Deprecated - use codeTransform_localBuildProject)",
```
justinmk3, line 3920:
```suggestion
            "description": "Deprecated - use codeTransform_localBuildProject",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -3476,9 +3553,58 @@
...
+                    "required": true
+                },
+                {
+                    "type": "reason",
+                    "required": false
+                }
+            ]
+        },
+        {
+            "name": "codeTransform_invokeApi",
```
justinmk3, line 3575: Please plan to deduplicate this metric in the clients. Emitting a metric for every service call is very noisy.

telemetry/definitions/commonDefinitions.json
```diff
@@ -297,6 +350,11 @@
             "type": "string",
             "description": "The ID of the job currently running"
         },
+        {
+            "name": "codeTransformLocalBuildSystemVersion",
```
justinmk3, line 354: I believe something like this was added in https://github.com/aws/aws-toolkit-common/pull/783

Please try to think of types that are more generic. There is no reason to have : 


```
codeTransformLocalBuildSystemVersion
someOtherFeatureLocalBuildSystemVersion
antherFeatureLocalBuildSystemVersion
```

Please, always apply this principle when designing metrics. Redundant, overlapping definitions makes it hard to find useful metrics, both when using them as well as visualizing them in the backend.

telemetry/definitions/commonDefinitions.json
tincheng: Would you recommend emitting API invocation metric only when the API call fails? The RequestId emitted in these API metrics have helped us find corresponding logs on code-whisperer side.

telemetry/definitions/commonDefinitions.json
```diff
@@ -248,10 +248,25 @@
             "type": "string",
             "description": "Any API-specific errors"
         },
+        {
+            "name": "codeTransformApiName",
```
justinmk3, line 252: If we are really going to start counting service calls (that really, really needs careful de-duplication treatment on the clients), then there is no reason for this field to be codetransform-specific.

We already have a "serviceType" field https://github.com/aws/aws-toolkit-common/blob/3b77ad1d8bd07708c0e4dc16c895f09cced41b0d/telemetry/definitions/commonDefinitions.json#L1546 

, so it's more discoverable to name this something like `serviceCall` or `serviceApiName`
```suggestion
            "name": "serviceApiName",
```

telemetry/definitions/commonDefinitions.json
tincheng: I discussed with @dhasani23 on #783 and we will include the same changes in this PR (and close that one). I'll change this to `localBuildSystemVersion`

telemetry/definitions/commonDefinitions.json
tincheng: Thinking more about this - it's true that emitting metrics for every service call has not been particularly helpful, and we do not evaluate those metrics. I'll remove the new changes related to `codeTransform_invokeApi`, but keep `codeTransform_logApi*` as deprecated.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3844,12 +3976,77 @@
...
+                },
+                {
+                    "type": "localBuildSystemVersion",
+                    "required": false
+                }
+            ]
+        },
+
+        {
+            "name": "codeTransform_uploadProject",
```
dhasani23, line 3988: If we're not emitting a metric for API calls, do we need to emit one for this PUT request? I'm indifferent but just wanted to mention it.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3863,7 +4060,7 @@
         },
         {
             "name": "codeTransform_vcsViewerCanceled",
-            "description": "An interactivity measured when users cancel a patch view.",
+            "description": "Deprecated - use codeTransform_viewDiff",
```
dhasani23, line 4063: codeTransform_viewDiff or codeTransform_viewArtifact?

telemetry/definitions/commonDefinitions.json
tincheng: This metric will be a step in user journey to indicate whether project has been uploaded successfully (including both zipping + upload).

telemetry/definitions/commonDefinitions.json
tincheng: Will fix!

telemetry/definitions/commonDefinitions.json
```diff
@@ -263,6 +263,39 @@
                 "ExportResultArchive"
             ]
         },
+        {
+            "name": "codeTransformArtifactType",
+            "type": "string",
+            "description": "Type of transform artifact",
+            "allowedValues": [
```
dhasani23, line 270: Should this also include something for the final diff.patch?

telemetry/definitions/commonDefinitions.json
```diff
@@ -251,7 +251,7 @@
         {
             "name": "codeTransformApiNames",
             "type": "string",
-            "description": "Names of allowed API calls",
+            "description": "Deprecated",
```
dhruvigajjar, line 254: please keep the description as it and add + deprecated. 
so that it's easy to lookup existing metrics.

tincheng: I went with the suggestion from @justinmk3 to keep description clear by leading with deprecated. I can add some original descriptions back for tracking purpose. 
```
"description": "Deprecated (Names of allowed API calls)"
```

telemetry/vscode/package-lock.json
tincheng, line 1: This is brought in by merge

telemetry/definitions/commonDefinitions.json
```diff
@@ -3430,23 +3473,50 @@
             "passive": true
         },
         {
-            "name": "codeTransform_buildComplete",
-            "passive": true,
-            "description": "During the transformation progress, log the status of the build.",
+            "name": "codeTransform_dependenciesCopied",
+            "description": "Deprecated - use codeTransform_uploadProject (The repo copies over at least one dependency successfully)",
```
dhruvigajjar, line 3477: could you also point to the new one along with deprecated

telemetry/definitions/commonDefinitions.json
```diff
@@ -3430,23 +3473,50 @@
...
+                    "type": "codeTransformSessionId",
                     "required": true
                 }
             ]
         },
         {
-            "name": "codeTransform_dependenciesCopied",
-            "description": "The repo copies over at least one dependency successfully.",
+            "name": "codeTransform_downloadArtifact",
+            "description": "Transform or user downloads code transform artifact",
```
dhruvigajjar, line 3487: When you transform or user downloads, do mean can be active or passive according to use case?

telemetry/definitions/commonDefinitions.json
```diff
@@ -3430,23 +3473,50 @@
...
+                    "type": "codeTransformTotalByteSize",
+                    "required": true
+                },
+                {
+                    "type": "reason",
+                    "required": false
+                },
+                {
+                    "type": "result",
+                    "required": true
```
dhruvigajjar, line 3519: please add `passive` if this metric is automatically generated.

telemetry/vscode/package-lock.json
```diff
@@ -3438,9 +3438,9 @@
             }
         },
         "node_modules/prettier": {
-            "version": "3.3.2",
```
justinmk3, line 3441: revert

justinmk3 (approved):

telemetry/definitions/commonDefinitions.json
tincheng: I think it will be better to change to user download, so I'll update it.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1309,6 +1348,11 @@
             "type": "int",
             "description": "Time from opening Composer to loading the file"
         },
+        {
+            "name": "localBuildSystemVersion",
```
justinmk3, line 1352: everything is assumed to be on the user's machine. use "remote" to call out when it's not. usually "local" is not necessary to call out.
```suggestion
            "name": "buildSystemVersion",
```

telemetry/definitions/commonDefinitions.json
tincheng: codeTransform_uploadProject is the new metric (that will contain a field to indicate whether dependencies are copied)

telemetry/definitions/commonDefinitions.json
tincheng: Right now it's not automatically generated (updated description)

telemetry/definitions/commonDefinitions.json
```diff
@@ -251,7 +256,7 @@
         {
             "name": "codeTransformApiNames",
             "type": "string",
-            "description": "Names of allowed API calls",
+            "description": "Deprecated (Names of allowed API calls)",
```
rli, line 259: ideally this results in deprecated annotations in the generated types, so that we can see this in the IDE

telemetry/definitions/commonDefinitions.json
```diff
@@ -167,6 +167,11 @@
             "type": "int",
             "description": "The amount of time required for the build to complete (in seconds)."
         },
+        {
+            "name": "buildSystemVersion",
+            "type": "string",
+            "description": "The build system version on the user's machine"
```
dhruvigajjar, line 173: can you make this more descriptive.

telemetry/definitions/commonDefinitions.json
```diff
@@ -263,6 +268,39 @@
...
+                "gradle",
+                "IDEBundledGradle"
+            ]
+        },
+        {
+            "name": "codeTransformBuildSystem",
+            "type": "string",
+            "description": "Type of build system",
+            "allowedValues": [
+                "Maven",
```
dhruvigajjar, line 300: is this related to `buildVersion`?

telemetry/definitions/commonDefinitions.json
dhasani23: We're keeping it generic because we support Maven and now Gradle

dhruvigajjar (approved):

dhruvigajjar (approved):

PR #794, opened 2024-07-22
telemetry: add version metadata to toolkit_initModule

farisab1:
Define version metadata on toolkit_initModule metric. This metadata will help us determine the corresponding version number of a user in each of the modules (ex: WebView2 version number in RuntimeInstaller module).

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -6351,6 +6351,10 @@
                 {
                     "type": "result",
                     "required": true
+                },
+                {
+                    "type": "version",
+                    "required": false
                 }
             ],
             "passive": true


telemetry/definitions/commonDefinitions.json
```diff
@@ -6351,6 +6351,10 @@
                 {
                     "type": "result",
                     "required": true
+                },
+                {
+                    "type": "version",
+                    "required": false
```
makenneth, line 6357: What would happen if this is set to `"required": true`? Seems like it would be an important field to have

makenneth (approved):

telemetry/definitions/commonDefinitions.json
rbbarad: `version` isn't an applicable field to all "modules"; it currently only has relevance to the "webview2" module. we wouldn't want this to be a required field for all modules

rbbarad (approved):

telemetry/definitions/commonDefinitions.json
farisab1: This would make it so that every module would need to specify this field. We left it as optional since currently only the RuntimeInstaller module has a version that it would pass through.

makenneth: Got it, thank you for the context!

dhruvigajjar (approved):

PR #797, opened 2024-07-26
Add new code transform metric fields and metric type for ApplyChanges

sunwoo-ms:
Updates `commonDefinitions.json` with some new fields used in existing `codeTransform` metric types, as well as a new metric type called `codeTransform_applyChanges` that is emitted whenever a user chooses to apply suggested changes.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -372,6 +372,11 @@
             "description": "The number of projects selected for transformation"
         },
         {
+            "name": "codeTransformNumberOfTransformedFiles",
+            "type": "int",
+            "description": "The number of files that were transformed during a transformation job"
+        },
+        {
             "name": "codeTransformNumberOfTransformedFilesAccepted",
             "type": "int",
             "description": "The number of transformed files that have been accepted by the user"
@@ -3633,6 +3638,10 @@
                 {
                     "type": "codeTransformJobId",
                     "required": true
+                },
+                {
+                    "type": "codeTransformNumberOfTransformedFiles",
+                    "required": false
                 },
                 {
                     "type": "codeTransformRuntimeError",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -249,9 +249,9 @@
             "description": "Time taken to invoke code scan service APIs in milliseconds"
         },
         {
-            "name": "codeTransformApiErrorMessage",
+            "name": "codeTransformErrorMessage",
             "type": "string",
-            "description": "Any API-specific errors"
+            "description": "The error message (can be local or API-side)"
         },
         {
             "name": "codeTransformApiNames",
@@ -367,9 +367,19 @@
             "description": "A general field for logging metadata associated to Amazon Q transform metrics."
         },
         {
-            "name": "codeTransformNumberOfProjects",
+            "name": "codeTransformNumberOfSelectedFiles",
+            "type": "int",
+            "description": "The number of files selected to apply changes to"
+        },
+        {
+            "name": "codeTransformNumberOfSelectedProjects",
             "type": "int",
             "description": "The number of projects selected for transformation"
+        },
+        {
+            "name": "codeTransformNumberOfSelectedUnits",
+            "type": "int",
+            "description": "The number of buildable units selected to apply changes to"
         },
         {
             "name": "codeTransformPatchViewerCancelSrcComponents",
@@ -3474,6 +3484,24 @@
             "passive": true
         },
         {
+            "name": "codeTransform_applyChanges",
+            "description": "Apply suggested changes to buildable units.",
+            "metadata": [
+                {
+                    "type": "codeTransformSessionId",
+                    "required": true
+                },
+                {
+                    "type": "codeTransformNumberOfSelectedUnits",
+                    "required": false
+                },
+                {
+                    "type": "codeTransformNumberOfSelectedFiles",
+                    "required": true
+                }
+            ]
+        },
+        {
             "name": "codeTransform_dependenciesCopied",
             "description": "Deprecated - use codeTransform_uploadProject (The repo copies over at least one dependency successfully)",
             "metadata": [
@@ -3716,7 +3744,7 @@
                     "required": false
                 },
                 {
-                    "type": "codeTransformNumberOfProjects",
+                    "type": "codeTransformNumberOfSelectedProjects",
                     "required": false
                 },
                 {
@@ -3741,6 +3769,10 @@
                 },
                 {
                     "type": "result",
+                    "required": false
+                },
+                {
+                    "type": "source",
                     "required": false
                 }
             ]
@@ -3816,10 +3848,10 @@
         {
             "name": "codeTransform_logApiError",
             "passive": true,
-            "description": "Deprecated - avoid logging every service call",
+            "description": "An API-side error has occurred.",
             "metadata": [
                 {
-                    "type": "codeTransformApiErrorMessage",
+                    "type": "codeTransformErrorMessage",
                     "required": true
                 },
                 {
@@ -3882,10 +3914,10 @@
         {
             "name": "codeTransform_logGeneralError",
             "passive": true,
-            "description": "Deprecated - avoid logging every service call",
+            "description": "A general error has occurred.",
             "metadata": [
                 {
-                    "type": "codeTransformApiErrorMessage",
+                    "type": "codeTransformErrorMessage",
                     "required": true
                 },
                 {
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -313,11 +313,6 @@
             "description": "If dependencies are copied into upload artifact."
         },
         {
-            "name": "codeTransformErrorMessage",
-            "type": "string",
-            "description": "The error message (can be local or API-side)"
-        },
-        {
             "name": "codeTransformJavaSourceVersionsAllowed",
             "type": "string",
             "description": "Allowed Java versions to transform from",
@@ -367,7 +362,7 @@
             "description": "A general field for logging metadata associated to Amazon Q transform metrics."
         },
         {
-            "name": "codeTransformNumberOfSelectedProjects",
+            "name": "codeTransformNumberOfProjects",
             "type": "int",
             "description": "The number of projects selected for transformation"
         },
@@ -3640,10 +3635,6 @@
                     "required": true
                 },
                 {
-                    "type": "codeTransformNumberOfTransformedFiles",
-                    "required": false
-                },
-                {
                     "type": "codeTransformRuntimeError",
                     "required": false
                 },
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -367,10 +367,20 @@
             "description": "A general field for logging metadata associated to Amazon Q transform metrics."
         },
         {
-            "name": "codeTransformNumberOfProjects",
+            "name": "codeTransformNumberOfSelectedFiles",
+            "type": "int",
+            "description": "The number of files selected to apply changes to"
+        },
```
jonlouie, line 373: This metric follows the same naming pattern as `codeTransformNumberOfSelectedProjects`, but the metrics themselves are very different (one is a pre-transform metric, the other is a post-transform metric). Can we use a more descriptive name that eliminates ambiguity? Suggestion: `codeTransformNumberOfTransformedFilesAccepted`.

Also, do we have a metric for # of transformed files? e.g. `codeTransformNumberOfTransformedFiles`

sunwoo-ms: This has been addressed in a more recent commit. Also, we now have a `codeTransformNumberOfTransformedFiles` field that is contained within the `codeTransform_downloadArtifact` metric.

jonlouie: Awesome, looks good.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3816,14 +3860,14 @@
         {
             "name": "codeTransform_logApiError",
             "passive": true,
-            "description": "Deprecated - avoid logging every service call",
+            "description": "An API-side error has occurred.",
             "metadata": [
                 {
-                    "type": "codeTransformApiErrorMessage",
+                    "type": "codeTransformApiNames",
```
dhruvigajjar, line 3866: why is ErrorMessage changed to APi names instead of `codeTransformApiErrorMessage`

telemetry/definitions/commonDefinitions.json
```diff
@@ -317,6 +312,11 @@
             "type": "boolean",
             "description": "If dependencies are copied into upload artifact."
         },
+        {
+            "name": "codeTransformErrorMessage",
```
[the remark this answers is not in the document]
dhruvigajjar: yes. please try to reuse types if possible. create your types only if you are exhautsing all existing error types. 
Also I would recommend creating a general type for errors instead of starting with a prefix

telemetry/definitions/commonDefinitions.json
```diff
@@ -367,10 +367,25 @@
...
             "type": "int",
             "description": "The number of projects selected for transformation"
         },
+        {
+            "name": "codeTransformNumberOfTransformedFiles",
+            "type": "int",
+            "description": "The number of files that were transformed during a transformation job"
+        },
+        {
+            "name": "codeTransformNumberOfTransformedFilesAccepted",
```
dhruvigajjar, line 380: https://github.com/aws/aws-toolkit-common/blob/21a260645545cb008992e47b768070353beaf18e/telemetry/definitions/commonDefinitions.json#L42-L51

are these the same fields? if your team is changing types, please make sure you clean the ones you dont reuse.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3716,7 +3756,7 @@
                     "required": false
                 },
                 {
-                    "type": "codeTransformNumberOfProjects",
+                    "type": "codeTransformNumberOfSelectedProjects",
```
dhruvigajjar, line 3759: this could be breaking change if it's already implemented in IDEs

telemetry/definitions/commonDefinitions.json
```diff
@@ -3816,14 +3860,14 @@
         {
             "name": "codeTransform_logApiError",
             "passive": true,
-            "description": "Deprecated - avoid logging every service call",
+            "description": "An API-side error has occurred.",
```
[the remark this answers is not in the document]
dhruvigajjar: Could you please sync with [Ting](tincheng) author of this PR. I think there's something out of sync here https://github.com/aws/aws-toolkit-common/pull/790/files

telemetry/definitions/commonDefinitions.json
tincheng: Is there a reason to capture these on the IDE side? Elastic Gumby backend already capture these metrics.

telemetry/definitions/commonDefinitions.json
```diff
@@ -367,10 +367,25 @@
             "description": "A general field for logging metadata associated to Amazon Q transform metrics."
         },
         {
-            "name": "codeTransformNumberOfProjects",
+            "name": "codeTransformNumberOfSelectedProjects",
```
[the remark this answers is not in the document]
tincheng: At the moment only one project can be transformed at a time for VSCode/IntelliJ, unless its different in Visual Studio? How is this metric used?

telemetry/definitions/commonDefinitions.json
```diff
@@ -3473,6 +3491,24 @@
             ],
             "passive": true
         },
+        {
+            "name": "codeTransform_applyChanges",
+            "description": "Apply suggested changes to buildable units.",
+            "metadata": [
```
tincheng, line 3497: I added a similar metric "codeTransform_viewArtifact" that can be emitted both when user accepted and rejected the changes through the field "userChoice". Maybe consider reusing that metrics, and feel free to add the new fields as non-required, since VSCode/IntelliJ will not need them.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3603,6 +3639,10 @@
                     "type": "codeTransformJobId",
                     "required": true
                 },
+                {
+                    "type": "codeTransformNumberOfTransformedFiles",
+                    "required": false
+                },
```
tincheng, line 3645: I think we should keep this metric's purpose clear: an user tries to initiate a transformation either through /transform or some other channel. codeTransformNumberOfTransformedFiles would belong more in codeTransform_jobStart per visual studio's usage. In addition, when user actually submits a transformation, there should be metric on the service side that captures the lines of code, number of files, etc, so consider using those metrics instead of introducing new ones on the IDE side.

telemetry/definitions/commonDefinitions.json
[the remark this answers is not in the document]
tincheng: Per my discussion with toolkit team (@dhruvigajjar and @justinmk3), we are aligned that IDE should not emit metrics for every single API calls. We should only emit metrics specific to transform status changes and user actions, hence I'm removing this metric. Please use `codeTransform_logGeneralError` if you need to emit some error metrics.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3882,10 +3926,10 @@
         {
             "name": "codeTransform_logGeneralError",
             "passive": true,
-            "description": "Deprecated - avoid logging every service call",
+            "description": "A general error has occurred.",
```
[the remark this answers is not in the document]
tincheng: We can keep this (remove the deprecated comment)

telemetry/definitions/commonDefinitions.json
jonlouie: This is exclusive to .NET Transform. We are essentially trying to track "out of the number of changes that the customer sees, how many do they accept?". This information can only be tracked on the IDE where the user interacts.

If changing the metrics names would help, do you have suggestions?

telemetry/definitions/commonDefinitions.json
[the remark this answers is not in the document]
sunwoo-ms: Okay, we've decided to keep this metric as being used and updated the document describing each metric type.

telemetry/definitions/commonDefinitions.json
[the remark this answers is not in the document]
sunwoo-ms: This has been addressed in a more recent commit. We're no longer un-deprecating this metric type.

telemetry/definitions/commonDefinitions.json
sunwoo-ms: This has been addressed in a more recent commit.

telemetry/definitions/commonDefinitions.json
sunwoo-ms: This has been addressed in a more recent commit. We now emit this field from the `codeTransform_jobStart` metric type instead.

telemetry/definitions/commonDefinitions.json
tincheng: Would your team be opened to use `codeTransform_downloadArtifact` instead of `codeTransform_jobArtifactDownloadAndDeserializeTime`? The metrics are for the same purpose, but the new one aligns with overall toolkit metric standard.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3818,14 +3857,6 @@
             "passive": true,
             "description": "Deprecated - avoid logging every service call",
             "metadata": [
-                {
-                    "type": "codeTransformApiErrorMessage",
-                    "required": true
-                },
```
dhruvigajjar, line 3824: are you sure this wont break any implementation?

telemetry/definitions/commonDefinitions.json
tincheng: Got it. I'm wondering if your team can reuse codeTransform_viewArtifact, which indicates users have interacted with the version control of the upgraded code. This can be added as a new (optional) field in it.

dhruvigajjar: are you planning to remove `amazonqNumberOfFilesGenerated` & `amazonqNumberOfFilesAccepted` repeating metrics?

telemetry/definitions/commonDefinitions.json
dhruvigajjar, line 1: can you list approximate frequencies of each metrics. we want to make sure we debounce most frequent metrics. Also it would be good idea to add results to each metric

telemetry/definitions/commonDefinitions.json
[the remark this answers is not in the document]
dhruvigajjar: Can you combine this metric to cover both success and failure case. I'm sure why we need a dedicated metric to cover failure

telemetry/definitions/commonDefinitions.json
sunwoo-ms: None of the other metric types make use of this field, and it was recommended by Chris to make use of one of the default fields in place of this one. But I can leave it if we'd rather not risk it.

telemetry/definitions/commonDefinitions.json
sunwoo-ms: Those fields are being utilized by metric types that we have not defined, so I haven't removed them in case that would be a breaking change.

telemetry/definitions/commonDefinitions.json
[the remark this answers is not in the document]
sunwoo-ms: There are errors that are either emitted by the API or that occur locally (e.g., build errors) that we want to be emitted so that we can track them for our operations.

dhruvigajjar (approved):

awschristou (approved):

PR #810, opened 2024-08-09
VSC telemetry updates

hayemaxi:
**Do not squash merge** Contains 2 commits: 

- feat: add additional fields to auth metrics

- feat(generation): allow awsRegion to be specified by callers in Metadata

Problem: `awsRegion` is a base metric and does not allow for callers to add it to their metadata when using `record()`. However, this field is often known best by the caller and not the telemetry guessing algorithm.

Solution: Update type generation to allow this base metric to be specified in metadata.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1590,6 +1590,16 @@
             "description": "The type of an SQS Queue"
         },
         {
+            "name": "ssoRegistrationClientId",
+            "type": "string",
+            "description": "The client ID of an SSO registration."
+        },
+        {
+            "name": "ssoRegistrationExpiresAt",
+            "type": "string",
+            "description": "Date/time that an SSO client registration expires."
+        },
+        {
             "name": "successCount",
             "type": "int",
             "description": "The number of successful operations"
@@ -2450,6 +2460,14 @@
                 },
                 {
                     "type": "source"
+                },
+                {
+                    "type": "ssoRegistrationClientId",
+                    "required": false
+                },
+                {
+                    "type": "ssoRegistrationExpiresAt",
+                    "required": false
                 }
             ]
         },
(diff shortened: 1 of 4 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1448,6 +1448,11 @@
             ],
             "description": "User-provided policy type of the reference file. Applicable to only CheckNoNewAccess check type."
         },
+        {
+            "name": "registrationExpiresAt",
```
justinmk3, line 1452: should this have a "sso" prefix like the other field added in this pr?
```suggestion
            "name": "ssoRegistrationExpiresAt",
```

telemetry/vscode/src/generate.ts
```diff
@@ -132,7 +132,10 @@ const commonMetadata = [
   'requestId',
   'requestServiceType',
   'result',
-]
+] as const
+
+/** Allowed to be manually specified in .record() calls */
+const allowedMetadata: typeof commonMetadata[number][] = ['awsRegion']
```
justinmk3, line 138: why not do this for all `commonMetadata`? what is the difference between commonMetadata vs allowedMetadata ?

telemetry/definitions/commonDefinitions.json
hayemaxi: Oops, this was intended, good catch

telemetry/vscode/src/generate.ts
justinmk3: ok, would `optionalMetadata` be a better name for this?

And a docstring on `commonMetadata`:

    /**
     * Fields set automatically by the telemetry client (thus application code
     * normally shouldn't set these because the value will be overridden).
     */

telemetry/vscode/src/generate.ts
```diff
@@ -132,7 +132,16 @@ const commonMetadata = [
...
+] as const
+
+/** 
+ * While all of the commonMetadata fields are managed by the telemetry service,
+ * i.e. automatically computed, some fields make sense to also be specified by
+ * the caller.
+ *
+ * These values are allowed to be specified in .record() calls.
+ */
+const allowedMetadata: typeof commonMetadata[number][] = ['awsRegion']
```
justinmk3, line 144: would `optionalMetadata` be a clearer name?

```suggestion
const optionalMetadata: typeof commonMetadata[number][] = ['awsRegion']
```

justinmk3 (approved):

PR #816, opened 2024-08-15
Add reasonDesc to createUpload definition

willyyhuang:
## Problem

As we emit `createUpload` event, we received multiple cases where the result is Failed but the reason is not enough for us to understand what exactly happened

## Solution

Added `reasonDesc` in `createUpload` definition, so we can add more information to debug customer issues.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1670,7 +1670,7 @@
                 },
                 {
                     "type": "reason",
-                    "required": false
+                    "required": false  
                 },
                 {
                     "type": "requestId",
@@ -1740,6 +1740,10 @@
                 {
                     "type": "reason",
                     "required": false
+                },
+                {
+                    "type": "reasonDesc",
+                    "required": false  
                 },
                 {
                     "type": "requestId",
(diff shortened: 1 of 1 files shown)


jpinkney-aws:
I thought reasonDesc already applied to every field: https://github.com/aws/aws-toolkit-common/blob/56866d841d58790c190f9a94fdb3b69b989761aa/telemetry/telemetryformat.md?plain=1#L77. Is that field not working for you?

willyyhuang:
> I thought reasonDesc already applied to every field:
> 
> https://github.com/aws/aws-toolkit-common/blob/56866d841d58790c190f9a94fdb3b69b989761aa/telemetry/telemetryformat.md?plain=1#L77
> 
> Is that field not working for you?

Yeah, using the generated Telemetry definition in Jetbrain, it doesn't seem to work, i haven't tried VSCode

awschristou:
`reasonDesc` is documented as being a "global" metadata field. It should be available for every metric.

https://github.com/aws/aws-toolkit-common/blob/main/telemetry/telemetryformat.md#global-arguments

justinmk3:
> Yeah, using the generated Telemetry definition in Jetbrain, it doesn't seem to work, i haven't tried VSCode

Please consider updating the jetbrains telememtry generator. This issue keeps coming up. Adding standard fields to every metric explicitly is a waste of time.

willyyhuang:
published: https://github.com/aws/aws-toolkit-common/pull/817

PR #823, opened 2024-08-21
feat: Add global flag for tracking runtime performance in a metric

jpinkney-aws:
## Problem
We should get some runtime information of q telemetry so that we can figure out what's going wrong through metrics, instead of customer reports

## Solution
I've added an optional flag that when enabled tells toolkits that they should start tracking performance information

## License


telemetry/telemetryformat.md
```diff
@@ -89,6 +89,8 @@ result: Result
 passive: boolean
 // Value based on unit and call type
 value: number
+// Indicates that the metric should track run-time performance information
+trackPerformance: boolean
```
awschristou, line 93: is it `performanceTracking` or `trackPerformance` ?

awschristou (commented):
I'd like this change to include updates to the C# and Java generators if possible. Would you like some help with that?

awschristou (approved):
Thanks for updating the generators!
Things look good to me for C#, please seek approval from @rli for java and @justinmk3 for typescript

telemetry/telemetrySchema.json
```diff
@@ -65,6 +65,10 @@
                     "passive": {
                         "type": "boolean",
                         "description": "Set as true if the metric is emitted without any explicit user action."
+                    },
+                    "trackPerformance": {
+                        "type": "boolean",
+                        "description": "Set as true if the metric can track run-time performance information."
```
justinmk3, line 71: should we mention "defaults to false" ?
```suggestion
                        "description": "Set as true if the metric can track run-time performance information. Default=false"
```

justinmk3 (approved):

rli:
please exclude jetbrains for now as the current design does not allow us to read the flag until it is too late

jpinkney-aws:
I've just dropped the jetbrains commit and added default=false to the telemetrySchema. Everything else stayed the same

PR #835, opened 2024-08-27
Add definitions for a feature's enabled state, and the opening and closing of a module

awschristou:
## Problem

We're revising the VS Toolkit's Getting Started experience to more cleanly enable and disable features. We wish to understand the user's journey through this experience, and identify when/why a feature gets enabled or disabled.

## Solution

A generalized set of metrics that can be applied to any feature:
- `toolkit_featureState` - intended to be emitted at various points in a workflow, to report whether or not a certain feature is enabled. Also emits "why" this state is being emitted (via the "source" field)
- `toolkit_openModule` - intended to be emitted when a feature/view/component of interest is opened/entered/displayed. "What" is being opened is represented by the "module" field, and "Why" it is being opened is represented in "source"
- `toolkit_closeModule` - intended to be emitted when a feature/view/component of interest is closed/exited. "What" is being opened is represented by the "module" field.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1200,7 +1200,9 @@
             "type": "string",
             "description": "The id of the feature the user is interacting in. See also: `component`, `featureId`, `module`, `resourceType`.",
             "allowedValues": [
+                "amazonQ",
                 "awsExplorer",
+                "awsToolkit",
                 "codewhisperer",
                 "codecatalyst"
             ]
@@ -6861,6 +6863,44 @@
             ]
         },
         {
+            "name": "toolkit_closeModule",
+            "description": "The user closed 'something' (specified by 'module'). Examples: a view, feature, resource, ...",
+            "metadata": [
+                {
+                    "type": "module",
+                    "required": true
+                }
+            ],
+            "passive": true
+        },
+        {
+            "name": "toolkit_featureState",
+            "description": "Represents the current enabled state of a feature. Used to track user journey through a feature. Emitted after feature-specific operations of interest in the Toolkit.",
+            "metadata": [
+                {
+                    "type": "authStatus",
+                    "required": false
+                },
+                {
+                    "type": "credentialSourceId",
+                    "required": false
+                },
+                {
+                    "type": "enabled",
+                    "required": true
+                },
+                {
+                    "type": "featureId",
+                    "required": true
+                },
+                {
+                    "type": "source",
+                    "required": true
+                }
+            ],
+            "passive": true
+        },
+        {
             "name": "toolkit_getExternalResource",
             "description": "The toolkit tried to retrieve blob data from a url",
             "metadata": [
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -6860,6 +6862,44 @@
...
+                {
+                    "type": "module",
+                    "required": true
+                }
+            ],
+            "passive": true
+        },
+        {
+            "name": "toolkit_featureState",
+            "description": "Represents the current enabled state of a feature. Used to track user journey through a feature. Emitted after feature-specific operations of interest in the Toolkit.",
```
justinmk3, line 6878: how does this differ from `aws_featureConfig` which was recently added? https://github.com/aws/aws-toolkit-common/blob/940ac45eeb76a5a6bc6b92c8a66306a0d1f363b7/telemetry/definitions/commonDefinitions.json#L2701C22-L2703

`aws_featureConfig` mentions AB testing but i don't see why it needs to be restricted to that. @ivikash

justinmk3: if this is truly different, would be helpful for the description to mention the differences

awschristou: I did consider using `aws_featureConfig`. My understanding of this metric is:
- it is a statement of immutable fact around a feature ("Toolkit has been told that for A/B purposes, feature X is A or B")
- it is likely not frequently emitted

The intent behind `toolkit_featureState` is:
- emitted often, representing the user's current state as they click or interact their way through a workflow
- the state is mutable (it is possible for User and/or Toolkit to affect the enabled state)
- indicates why state is being emitted ("User has specified that feature is enabled" or "User is entering this workflow, here's the starting value")

Based on the requirements I discussed with @dhruvigajjar , `aws_featureConfig` did not match the needs.

justinmk3: > representing the user's current state as they click or interact their way through a workflow

This "journey through a feature" sounds like it could live on the existing feature-related metrics. Seems like we're inverting things with `toolkit_featureState`: now the "feature" will live as a field on `toolkit_featureState`, even though each "feature" already has dedicated metrics (which could themselves have a `featureId` and related fields).

Feels like a denormalization

awschristou: I think that's okay. If we had done this in the past, we may have used it more. I'm not inclined to make metrics with "gettingStarted" in the name, that is how we continue to proliferate metrics definitions. Not that that is a bad thing, but fewer is easier to manage.

dhruvigajjar: `featureConfig` has lot of fields which are only specific to AB testing which might be confusing during interpretation.
Since VS's getting started page has purpose of giving user feature specific (AWS Toolkit or AmazonQ)experience I thought dedicated metric for user's experience is `enabled` or not would be helpful, thus `toolkit_featureState`. 

I see this as equivalent of VSC/JB's `auth_addConnection` (as part of getting started page/auth page) that gives information about user's end-to-end experience through auth flow,`toolkit_featureState` will give information about user's journey through enabling a feature.

justinmk3 (approved):
This could be addressed by "container metrics": https://sim.amazon.com/issues/IDE-10178

oh well

dhruvigajjar (approved):

PR #836, opened 2024-08-27
telemetry(amazonq): make fields/metrics more generic to handle Gradle

dhasani23:
## Problem

Some fields/metrics were too Maven-specific, and we have Gradle support incoming.

## Solution

Make said fields/metrics more generic.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -404,8 +404,8 @@
                 "NoJavaProject",
                 "MixedLanguages",
                 "UnsupportedJavaVersion",
-                "ProjectJDKDiffersFromMavenJDK",
-                "NonMavenProject",
+                "ProjectJDKDiffersFromBuildSystemJDK",
+                "UnsupportedBuildSystem",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -4249,7 +4249,7 @@
                 },
                 {
                     "type": "codeTransformBuildSystem",
-                    "required": true
+                    "required": false
```
dhasani23, line 4252: This should be optional because we use `validateProject` to emit cases where the user has no projects open, no Maven/Gradle projects open, no Java projects open, etc. and at that point the build system has not been set/determined yet.

telemetry/definitions/commonDefinitions.json
```diff
@@ -4166,7 +4166,7 @@
                 },
                 {
                     "type": "userChoice",
-                    "required": true
+                    "required": false
```
dhasani23, line 4169: Also made this optional because there are some cases where it's not really applicable

justinmk3 (approved):

PR #839, opened 2024-08-30
Add definitions for a webview being loaded/erroring

jpinkney-aws:
## Problem
Port webview_load/webview_error metrics from vscode to commons.

- `webview_load` is called when a webview is loaded
- `webview_error` is called when an error occurs in the webview

### I'm trying to answer the following questions with these metrics :

    How long are webviews taking to load?
    How long until they're interactive?
    What errors are occurring in the webview?

## License


telemetry/definitions/commonDefinitions.json
```diff
@@ -1704,6 +1704,11 @@
             "type": "string",
             "description": "A generic version metadata"
         },
+        {
+            "name": "webviewName",
```
nkomonen-amazon, line 1708: We already have some existing types such as `name` and `id` which we should resuse. The idea is that these are generalized types that mean different things in the context of the metric that they are in.

jpinkney-aws: Didn't know those existed. Apparently you need to search by `: "name"` so it doesn't get clobbered together with the other type names

nkomonen-amazon (approved):

awschristou:
I suggest building on toolkit_openModule  from #835 and making these 
- toolkit_loadModule
- toolkit_errorModule

These become more generalized than "web views"

telemetry/definitions/commonDefinitions.json
```diff
@@ -7138,6 +7138,26 @@
                     "type": "result"
                 }
             ]
+        },
+        {
+            "name": "webview_error",
```
justinmk3, line 7143: Why is this needed? Every event can store `result`, `reason`, `reasonDesc`. Errors should not be modeled as separate events, they should be modeled as existing events with `result=Failed` + `reason`.

E.g. in this case webview_load could be such an event. If failure can occur after load, and we want a generic "webview interaction" event, then let's consider that instead. E.g. `webview_click`, `webview_submit`, etc.

justinmk3:
> These become more generalized than "web views"

+1

telemetry/definitions/commonDefinitions.json
jpinkney-aws: webview_error in this case is an generic error that can get thrown inside of mynah ui. Not sure if you can tie it directly to a chatMessage or anything since it can appear whenever: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/amazonq/webview/ui/main.ts#L26

jpinkney-aws:
Does `Called when an unexpected error occurs within a module outside of regular metrics tracking` make it more clear where `toolkit_errorModule` can be used?

In VSCode it's goal is to track: `window.addEventListener('error')` messages that can occur at any time for any reason

justinmk3:
> Does `Called when an unexpected error occurs within a module outside of regular metrics tracking` make it more clear where `toolkit_errorModule` can be used?
> 
> In VSCode it's goal is to track: `window.addEventListener('error')` messages that can occur at any time for any reason

That could make sense. Mentioning "Example: window.addEventListener('error')" in the description could help too.

OTOH, if we want a generic thing that captures unknown webview events, can we not name it `webview_event` ? That leaves the door open for capturing other events (both successes and failures). This is similar to `ui_click` and `vscode_executeCommand` which are rather generic but capture both success and failure.

telemetry/definitions/commonDefinitions.json
justinmk3: see https://github.com/aws/aws-toolkit-common/pull/839#issuecomment-2332486730

PR #856, opened 2024-09-18
telemetry: toolkit_trackScenario

tverney:
## Problem
No generic metric for counting how many times a (potentially temporary) scenario is reached.
Such scenarios may be temporary and thus a dedicated metric is not appropriate.

Example (vscode):
When customer use workspaces to generate files, RTS and LLM cannot decide which folder to deliver the generated file. We want track how many times we reach for accountability and prioritization.

## Solution
Add toolkit_trackScenario.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1897,25 +1897,6 @@
             ]
         },
         {
-            "name": "amazonq_trackScenarioCountUsage",
-            "description": "Count how many times a scenario is reached.",
-            "unit": "Count",
-            "metadata": [
-                {
-                    "type": "amazonqConversationId",
-                    "required": true
-                },
-                {
-                    "type": "scenarioType",
-                    "required": true
-                },
-                {
-                    "type": "credentialStartUrl",
-                    "required": false
-                }
-            ]
-        },
-        {
             "name": "amazonq_codeGenerationThumbsDown",
             "description": "User clicked on the thumbs down button to say that they are unsatisfied",
             "unit": "Count",
@@ -2162,6 +2143,25 @@
                 },
                 {
                     "type": "result",
+                    "required": false
+                }
+            ]
+        },
+        {
+            "name": "amazonq_trackScenarioCountUsage",
+            "description": "Count how many times a scenario is reached.",
+            "unit": "Count",
+            "metadata": [
+                {
+                    "type": "amazonqConversationId",
+                    "required": true
+                },
+                {
+                    "type": "scenarioType",
+                    "required": true
+                },
+                {
+                    "type": "credentialStartUrl",
                     "required": false
                 }
             ]
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -6817,12 +6817,12 @@
                     "type": "amazonqConversationId",
                     "required": true
                 },
+                 {
+                    "type": "count"
+                },
                 {
                     "type": "credentialStartUrl",
                     "required": false
-                },
-                 {
-                    "type": "count"
                 },
                 {
                     "type": "scenario",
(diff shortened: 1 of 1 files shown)


justinmk3:
> suggested adding a new generic metric because we currently don't have one

Ok. But `amazonq_trackScenarioCountUsage` is rather specific. Why does it mention "count" in the name? If we want to attach other fields (besides the "count") for that scenario, then we would need a new metric... 

why not name it `amazonq_trackScenario` ? But also, every metric tracks a "scenario" already, so adding this metric which has a `scenario` field seems questionable.

tverney:
> > suggested adding a new generic metric because we currently don't have one
> 
> Ok. But `amazonq_trackScenarioCountUsage` is rather specific. Why does it mention "count" in the name? If we want to attach other fields (besides the "count") for that scenario, then we would need a new metric...
> 
> why not name it `amazonq_trackScenario` ? But also, every metric tracks a "scenario" already, so adding this metric which has a `scenario` field seems questionable.

Well, the count its because we want limit the telemetry to just count the number of times we reach a specific condition. It was specific before, however, it was suggested to amplify to something more generic, so now it's talking about a "scenario". We will not be doing more than just counting for the moment. If you like it better, we could call it amazonq_trackConditionCountUsage. I'm ok with either naming conventions its good for you, the goal here its just have metrics on a "specific" condition for us.

justinmk3:
> We will not be doing more than just counting for the moment.

But that's the point: we need to think beyond jus this moment. 

I suggest naming the metric `toolkit_trackScenario`. This metric will be flexible enough to be used for various purposes that aren't well-defined. Later, when a scenario becomes well-defined, then it can and should have a well-defined metric.

telemetry/definitions/commonDefinitions.json
```diff
@@ -2146,6 +2154,25 @@
                 }
             ]
         },
+        {
+            "name": "amazonq_trackScenarioCountUsage",
+            "description": "Count how many times a scenario is reached.",
+            "unit": "Count",
```
justinmk3, line 2160: This is fine, and/or you can add the `count` field in the `metadata` list below. https://github.com/aws/aws-toolkit-common/blob/dff810a4bc80a7f10b22951fc7517ec5ff6334a2/telemetry/definitions/commonDefinitions.json#L887

tverney:
> > We will not be doing more than just counting for the moment.
> 
> 
> 
> But that's the point: we need to think beyond just this moment. 
> 
> 
> 
> I suggest naming the metric `toolkit_trackScenario`. You can add any fields you want in the metric. Since you only want "count" for now, just add the existing `count` field: https://github.com/aws/aws-toolkit-common/blob/dff810a4bc80a7f10b22951fc7517ec5ff6334a2/telemetry/definitions/commonDefinitions.json#L887
> 
> 
> 
> This metric will be flexible enough to be used for various purposes that aren't well-defined. Later, when a scenario becomes well-defined, then it can and should have a well-defined metric.

Ok,SGTM. I'll update accordingly.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1560,6 +1560,14 @@
             ],
             "description": "Type of save executed"
         },
+        {
+            "name": "scenarioType",
```
justinmk3, line 1564: all types are types. don't need "type" in the name.

```suggestion
            "name": "scenario",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -1560,6 +1560,14 @@
...
+            "type": "string",
+            "allowedValues": [
```
justinmk3, line 1566: shall we define this as a string instead of enum? the purpose of this metric is to be low-friction so a free-form string may be better.

telemetry/definitions/commonDefinitions.json
```diff
@@ -6801,6 +6806,30 @@
             ],
             "passive": true
         },
+
+        {
+            "name": "toolkit_trackScenario",
+            "description": "Count how many times a scenario is reached.",
```
justinmk3, line 6812:
```suggestion
            "description": "Generic metric for tracking arbitrary scenarios that are not yet formalized into a full metric.",
```

PR #859, opened 2024-09-20
refactor(vscode): remove telemetry.foo.record

jpinkney-aws:
## Problem
In vscode `telemetry.foo.record` is confusing to partner teams and has a couple of misuses in our codebase

## Solution
We're modifying vscodes telemetry generation to only allow records on a span, instead of on `telemetry.foo.record`.

This means the following will be available
```
telemetry.foo.run(span => {
    span.record( {someInformation: 5})
})
```

but `telemetry.foo.record` will not.

## License

### DIFF
--- a/telemetry/vscode/src/generate.ts
+++ b/telemetry/vscode/src/generate.ts
@@ -402,6 +402,7 @@ function generateFile(telemetryJson: MetricDefinitionRoot, dest: string) {
         name: 'Span',
         isExported: true,
         typeParameters: [{ name: 'T' }],
+        docs: ['Represents a telemetry span for tracking and recording metric data.'],
         methods: [
             {
                 name: 'record',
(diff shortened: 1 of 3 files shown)


telemetry/vscode/src/generate.ts
```diff
@@ -406,10 +397,30 @@ function generateFile(telemetryJson: MetricDefinitionRoot, dest: string) {
         kind: StructureKind.TypeAlias,
     }
 
+    const span: InterfaceDeclarationStructure = {
+        kind: StructureKind.Interface,
```
justinmk3, line 401: is there a way to add `docs` so the generated result will have a docstring ?

justinmk3 (approved):
Thank you!

telemetry/vscode/src/generate.ts
jpinkney-aws: Apparently its possible, I just added "'Represents a telemetry span for tracking and recording metric data.'"

PR #885, opened 2024-10-16
telemetry: add `remoteWorkspace` connect type for ec2. 

Hweinstock:
## Problem
Part of the work here: https://github.com/aws/aws-toolkit-vscode/pull/3723
## Solution
See PR above. 

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1131,7 +1131,8 @@
                 "remoteDesktop",
                 "ssh",
                 "scp",
-                "ssm"
+                "ssm", 
+                "remoteWorkspace"
             ],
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1131,7 +1131,8 @@
                 "remoteDesktop",
                 "ssh",
                 "scp",
-                "ssm"
+                "ssm", 
+                "remoteWorkspace"
             ],
             "description": "Ways to connect to an EC2 Instance"
```
justinmk3, line 1137: Just for reference: this description is where we can explain certain fields, if we want to. E.g. in the future if someone wonders what the difference is between remoteDesktop vs remoteWorkspace, we could answer that here.

PR #817, opened 2024-08-16
feat(jetbrains): Add global metadata fields into all metrics for jetbrains telemetry generator

willyyhuang:
## Problem

Jetbrain telemetry definition does not contain global metadata fields.

## Solution

Added global metadata fields into every metrics in telemetry generator

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #890, opened 2024-10-22
telemetry: add definitions for reviewbird

ctlai95:
## Problem

Missing telemetry events for reviewbird changes

## Solution

Add definitions for reviewbird changes

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -4427,6 +4427,37 @@
             ]
         },
         {
+            "name": "codewhisperer_codeScanIssueGenerateFix",
+            "description": "Called when a code scan issue fix is generated",
+            "metadata": [
+                {
+                    "type": "component"
+                },
+                {
+                    "type": "credentialStartUrl",
+                    "required": false
+                },
+                {
+                    "type": "detectorId"
+                },
+                {
+                    "type": "findingId"
+                },
+                {
+                    "type": "reason",
+                    "required": false
+                },
+                {
+                    "type": "result",
+                    "required": true
+                },
+                {
+                    "type": "ruleId",
+                    "required": false
+                }
+            ]
+        },
+        {
             "name": "codewhisperer_codeScanIssueHover",
             "description": "Called when a code scan issue is hovered over",
             "metadata": [
@@ -4442,26 +4473,6 @@
                 },
                 {
                     "type": "includesFix"
-                },
-                {
-                    "type": "ruleId",
-                    "required": false
-                }
-            ]
-        },
-        {
-            "name": "codewhisperer_codeScanIssueViewDetails",
-            "description": "Called when a code scan issue webview is opened",
-            "metadata": [
-                {
-                    "type": "credentialStartUrl",
-                    "required": false
-                },
-                {
-                    "type": "detectorId"
-                },
-                {
-                    "type": "findingId"
                 },
                 {
                     "type": "ruleId",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -4322,6 +4330,72 @@
             ],
             "passive": true
         },
+        {
+            "name": "codewhisperer_codeScanChatFileScan",
+            "description": "Called when file scan is selected from chat",
+            "metadata": [
+                {
+                    "type": "codeScanSessionId",
```
justinmk3, line 4338: does this map to a service-side id? FYI, every `telemetry.xx.run(...)` "tree" of metrics automatically shares a common `traceId`: https://github.com/aws/aws-toolkit-vscode/pull/5639

telemetry/definitions/commonDefinitions.json
```diff
@@ -4322,6 +4330,72 @@
...
+                {
+                    "type": "credentialSourceId",
+                    "required": false
+                },
+                {
+                    "type": "reason",
+                    "required": false
+                },
+                {
+                    "type": "result",
```
justinmk3, line 4372: these are redundant and should not be explicitly added. all toolkits now support these fields automatically, see https://github.com/aws/aws-toolkit-common/pull/817

justinmk3, line 4372: `reason`, `result`, and other fields listed [in the docs](https://github.com/aws/aws-toolkit-common/blob/main/telemetry/telemetryformat.md#global-arguments) should not be explicitly added. all toolkits now support these fields automatically, see https://github.com/aws/aws-toolkit-common/pull/817

telemetry/definitions/commonDefinitions.json
```diff
@@ -4322,6 +4330,48 @@
...
+                },
+                {
+                    "type": "credentialSourceId",
+                    "required": false
+                }
+            ]
+        },
+        {
+            "name": "codewhisperer_codeScanChatProjectScan",
+            "description": "Called when project scan is selected from chat",
```
justinmk3, line 4363: this metric and `codewhisperer_codeScanIssueIgnore` look over-specific. it's not maintainable or scalable to create a metric for every tiny variation. instead, consider a generalized metric like

```
codewhisperer_codeScanScan {
  source = 'chat' | ...
  type = 'file' | 'project'
}
```

If that's not acceptable, at least avoid baking `Chat` in the metric name. It's not unlikely that other "sources" will trigger a scan in the future, and it is again costly to define a metric for each and every scenario, which a more generic metric could instead be parameterized. That also helps dashboard maintenance...

telemetry/definitions/commonDefinitions.json
ctlai95: Got it, I will remove this

telemetry/definitions/commonDefinitions.json
ctlai95: Makes sense, updating

telemetry/definitions/commonDefinitions.json
```diff
@@ -4375,6 +4441,75 @@
...
+                },
+                {
+                    "type": "ruleId",
+                    "required": false
+                }
+            ]
+        },
+        {
+            "name": "codewhisperer_codeScanIssueRegenerateFix",
+            "description": "Called when a code scan issue fix is regenerated",
```
justinmk3, line 4492: why does this need to be a separate metric vs `codewhisperer_codeScanIssueGenerateFix`, instead of a flag

PR #905, opened 2024-11-10
telemetry: Metrics for amazonq unit test generation

laileni-aws:
## Problem
- No metrics for amazonq unit test generation
## Solution
Planning to add 4 metrics for Unit test generation.
1. amazonq_unitTestGeneration(new)
2. amazonq_utg_generateTests(new)
3. amazonq_utg_buildLoop(new)
4. ui_click(existing)

- Planning to add a new optional field (step) to ui_click to get metrics on which stage user has clicked a button.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1713,7 +1713,7 @@
         {
             "name": "isSupportedLanguage",
             "type": "boolean",
-            "description": "supported language"
+            "description": "Indicate if the language is supported"
         },
         {
             "name": "syncedResources",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1736,6 +1736,11 @@
             "description": "Date/time that an SSO client registration expires."
         },
         {
+            "name": "step",
+            "type": "string",
+            "description": "Indicates the stage at which a user interface click action was performed." 
+        },
+        {
             "name": "successCount",
             "type": "int",
             "description": "The number of successful operations"
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1433,6 +1433,11 @@
             "description": "The Lambda Package type of the function"
         },
         {
+            "name": "language",
+            "type": "string",
+            "description": "Language used for the project."
+        },
+        {
             "name": "latencyOfClientSideBuild",
             "type": "int",
             "description": "Time from start to end of Client-side build."
@@ -1441,11 +1446,6 @@
             "name": "latencyOfGeneratingTests",
             "type": "int",
             "description": "Time from start to end of Generating unit tests"
-        },
-        {
-            "name": "language",
-            "type": "string",
-            "description": "Language used for the project."
         },
         {
             "name": "linesOfCodeGenerated",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2350,6 +2350,10 @@
                     "required": false
                 },
                 {
+                    "type": "numberOfBuildPassed",
+                    "required": false
+                },
+                {
                     "type": "numberOfUnitTestCasesAccepted",
                     "required": false
                 },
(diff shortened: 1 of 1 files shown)


ashishrp-aws (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -1457,6 +1497,31 @@
...
+        {
+            "name": "numberOfBuildFailed",
+            "type": "int",
+            "description": "Number of build failures in unit test generation build loop"
+        },
+        {
+            "name": "numberOfBuildPassed",
+            "type": "int",
+            "description": "Number of build succeeded in unit test generation build loop"
+        },
```
justinmk3, line 1514: there are already some generic "successCount" and similar fields. use those. if you need to add a `executedCount` or something like that, that is preferred.

do not add over-specific fields for every metric.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1457,6 +1497,31 @@
...
+        {
+            "name": "numberOfUnitTestCasesAccepted",
+            "type": "int",
+            "description": "Number of unit test cases accepted"
+        },
+        {
+            "name": "numberOfUnitTestCasesGenerated",
+            "type": "int",
```
justinmk3, line 1522: use or create generic `fooCount` metrics.

please always do this in the future. I have mentiond this in previous reviews.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1670,11 +1735,21 @@
...
         {
             "name": "successCount",
             "type": "int",
             "description": "The number of successful operations"
         },
+        {
+            "name": "supportedLanguage",
+            "type": "boolean",
+            "description": "supported language"
+        },
```
justinmk3, line 1752: no.

telemetry/definitions/commonDefinitions.json
laileni-aws: Named these two fields to common`acceptedCount` and `generatedCount`.

telemetry/definitions/commonDefinitions.json
laileni-aws: I didn't see a field which checks the supported language(boolean).
LMK If I am missing something.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1694,6 +1749,16 @@
             "type": "string",
             "description": "Generic name of a template"
         },
+        {
+            "name": "testGenerationJobGroupName",
+            "type": "string",
+            "description": "Unit test generation job group name"
+        },
+        {
+            "name": "testGenerationJobId",
```
justinmk3, line 1758: a group name and job id could be used in many metrics. please do not add hyper-specific fields for every metric.

I already see fields like `codeTransformJobId`, and that's another case where we're paying the price for those fields being over-specific.

```suggestion
            "name": "jobGroup",
            "type": "string",
            "description": "Job group name associated with the metric"
        },
        {
            "name": "jobId",
```

telemetry/definitions/commonDefinitions.json
chandankmishra: This field should be isSupportedLanguage to indicate it is boolean field. SupportedLanguage seems like we are going to pass the actual value of language which could be string/enum type.

Also the comment should be "Indicate if the language is supported"

telemetry/definitions/commonDefinitions.json
```diff
@@ -1417,6 +1442,26 @@
...
+            "type": "int",
+            "description": "Time from start to end of Generating unit tests"
+        },
+        {
+            "name": "linesOfCodeAccepted",
+            "type": "int",
+            "description": "Number of lines of code accepted"
+        },
+        {
+            "name": "linesOfCodeGenerated",
```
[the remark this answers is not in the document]
laileni-aws: They are using `cwsprChatAcceptedNumberOfLines` which is specific name not generic but trying to use the same.

telemetry/definitions/commonDefinitions.json
laileni-aws: Will update with this

telemetry/definitions/commonDefinitions.json
laileni-aws: I did not see convention naming isXXXX for boolean values in the json file so for keeping consistent went with `supportedLanguage`
Let me change this accordingly

PR #913, opened 2024-11-13
feat: add general latency metrics

jpinkney-aws:
## Problem
- Soooooooooo many teams have specific latency metric fields and many are asking to add them for re-invent

## Solution
- De-dup a lot of them e.g. amazonqGenerateApproachLatency, amazonqGenerateCodeResponseLatency, amazonqEndOfTheConversationLatency, etc

## Other notes
- amazonq_approachInvoke would then have perfServerLatency instead of an individual amazonqGenerateApproachLatency field
- amazonq_codeGenerationInvoke would then have perfServerLatency instead of an individual amazonqGenerateCodeResponseLatency field
- amazonq_endChat would then have perfE2ELatency instead of an individual amazonqEndOfTheConversationLatency field

- In the future we can re-use the perfClientLatency metric for tracking the duration of different events that only happen in the client side (e.g. how long the client latency takes in the amazon q chat flow)

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -210,6 +210,11 @@
             "description": "User inputted check type to denote which custom check to run."
         },
         {
+            "name": "clientLatency",
+            "type": "double",
+            "description": "The time duration in milliseconds to process an action on the client side"
+        },
+        {
             "name": "cloudWatchLogsPresentation",
             "type": "string",
             "allowedValues": [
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -209,6 +209,11 @@
             ],
             "description": "User inputted check type to denote which custom check to run."
         },
+        {
+            "name": "clientLatency",
```
justinmk3, line 213: Since teams have indicated that it's hard to find related metrics, should we deviate a bit from our normal pattern and use `latency` as a prefix for all of this? Or maybe even `perf` as a prefix would help group all perf-related fields in the future.

awschristou:
It might be worth illustrating in the PR description how some existing metrics would be adjusted to use these proposed fields. This also helps to weed out if there are metrics where your proposal wouldn't be suitable. Other than that, I like this idea.

telemetry/definitions/commonDefinitions.json
jpinkney-aws: I think things like latency and duration are already naturally searchable. What advantages would we get if we prefixed with latency or perf? My problem with telemetry right now is every time I search for something that could be general it starts with the prefix of a team/project name 😅

justinmk3: they will be grouped together

PR #916, opened 2024-11-19
feat(cwl): Add metric definitions for LiveTail

keeganirby:
## Problem
Cloudwatch wants to monitor usage metrics of the LiveTail integration

## Solution
When starting a session emit telemetry for: 
* Session already started (this case happens when session is already running, and customer sends a new command that matches the already running session)
* Has LogEventFilter
* LogStream filter type
* Source of the command (command palette, explorer)

When closing a session: 
* Session duration
* source of cancellation (ex: CodeLens, ClosingEditors)

---

License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -5023,6 +5023,9 @@
                     "required": false
                 },
                 {
+                    "type": "result"
+                },
+                {
                     "type": "sessionAlreadyStarted",
                     "required": true
                 },
@@ -5039,6 +5042,9 @@
                 {
                     "type": "duration",
                     "required": true
+                },
+                {
+                    "type": "result"
                 },
                 {
                     "type": "source",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1782,6 +1782,21 @@
             "name": "xrayEnabled",
             "type": "boolean",
             "description": "Whether or not AWS X-Ray is enabled"
+        },
+        {
+            "name": "sessionAlreadyStarted",
+            "type": "boolean",
+            "description": "Session already open"
+        },
+        {
+            "name": "hasLogEventFilterPattern",
+            "type": "boolean",
+            "description": "If LogEvent filter pattern is applied"
+        },
+        {
+            "name": "logStreamFilterType",
+            "type": "string",
+            "description": "Type of LogStream filter applied to session"
         }
     ],
     "metrics": [
@@ -7080,6 +7095,42 @@
             "metadata": [
                 {
                     "type": "result"
+                }
+            ]
+        },
+        {
+            "name": "cwlLiveTail_Start",
+            "description": "When user starts a new LiveTail command",
+            "metadata": [
+                {
+                    "type": "source",
+                    "required": true
+                },
+                {
+                    "type": "sessionAlreadyStarted",
+                    "required": true
+                },
+                {
+                    "type": "hasLogEventFilterPattern",
+                    "required": false
+                },
+                {
+                    "type": "logStreamFilterType",
+                    "required": false
+                }
+            ]
+        },
+        {
+            "name": "cwlLiveTail_Stop",
+            "description": "When user stops a liveTailSession",
+            "metadata": [
+                {
+                    "type": "source",
+                    "required": true
+                },
+                {
+                    "type": "duration",
+                    "required": true
                 }
             ]
         }
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1298,6 +1298,11 @@
             "description": "Whether the user has access to CodeWhisperer Chat"
         },
         {
+            "name": "hasLogEventFilterPattern",
+            "type": "boolean",
+            "description": "If LogEvent filter pattern is applied"
+        },
+        {
             "name": "hasTextFilter",
             "type": "boolean",
             "description": "A text based filter was used"
@@ -1437,6 +1442,11 @@
             "name": "locale",
             "type": "string",
             "description": "User locale. Examples: en-US, en-GB, etc."
+        },
+        {
+            "name": "logStreamFilterType",
+            "type": "string",
+            "description": "Type of LogStream filter applied to session"
         },
         {
             "name": "metricId",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -4995,6 +5010,42 @@
...
+                    "type": "logStreamFilterType",
+                    "required": false
+                },
+                {
+                    "type": "sessionAlreadyStarted",
+                    "required": true
+                },
+                {
+                    "type": "source",
+                    "required": true
```
dhruvigajjar, line 5031: can you add result to `cwlLiveTail_Start` that suggests what happened as result of that command? and if applicabl capture error reason for Failed cases.

telemetry/definitions/commonDefinitions.json
```diff
@@ -4995,6 +5010,42 @@
...
+            "metadata": [
+                {
+                    "type": "duration",
+                    "required": true
+                },
+                {
+                    "type": "source",
+                    "required": true
+                }
+            ]
```
dhruvigajjar, line 5047: same comment for result.

dhruvigajjar (approved):

telemetry/definitions/commonDefinitions.json
```diff
@@ -1297,6 +1297,11 @@
             "type": "boolean",
             "description": "Whether the user has access to CodeWhisperer Chat"
         },
+        {
+            "name": "hasLogEventFilterPattern",
```
justinmk3, line 1301: This is over-specific. Can you find a field name that will be useful for other metrics in the future, e.g. `hasPatternFilter` (why is "log event" important enough to call out in the field name?)



Another way of future-proofing is to use enum-like (or string) fields instead of booleans. It's pretty cumbersome that we have booleans for each of these variants, and may need to add even more in the future.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1297,6 +1297,11 @@
...
+            "type": "boolean",
+            "description": "If LogEvent filter pattern is applied"
+        },
         {
             "name": "hasTextFilter",
```
justinmk3, line 1306: How is `hasLogEventFilterPattern` different than this existing field?

telemetry/definitions/commonDefinitions.json
```diff
@@ -4995,6 +5010,48 @@
                 }
             ]
         },
+        {
+            "name": "cwlLiveTail_Start",
```
justinmk3, line 5014: I am certain that we have existing cloudwatch metrics. Yet these new "live tail" metrics don't use the same prefix, so now both will be harder to discover.

telemetry/definitions/commonDefinitions.json
dhruvigajjar: yes I missed _hasTextFilter_ which in itself is over specified and could have been _filterType_ with enum of text, time, log event etc. but at times metrics requires multiple fields each with it's corresponding boolean values. such has CW logs could have both time & text filter and hence we would go with two fields. 

In this case, I agree we could reuse _hasTextFilter_. @keeganirby could you please reuse _hasTextFilter_ in codebase and then we can safely remove _hasLogEventFilterPattern_

telemetry/definitions/commonDefinitions.json
dhruvigajjar: my bad. I perceived this as a service on it's own. could we rename it to `cloudwatchlogs_Start`. 
if needed later we could add field distinguishing as `type:liveTail`

justinmk3: > could we rename it to `cloudwatchlogs_Start`.

I think `cloudwatchlogs_startLiveTail` would be fine. As long as the same prefix is used as other CWL metrics.

telemetry/definitions/commonDefinitions.json
keeganirby: Yeah we can use hasTextFilter.

telemetry/definitions/commonDefinitions.json
```diff
@@ -1438,6 +1443,11 @@
             "type": "string",
             "description": "User locale. Examples: en-US, en-GB, etc."
         },
+        {
+            "name": "logStreamFilterType",
+            "type": "string",
+            "description": "Type of LogStream filter applied to session"
```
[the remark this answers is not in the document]
keeganirby: We have two different filter parameters:
* logStream: determines from which logStream(s) log events are included
* LogEvent: only includes logEvents that match a specified pattern. 

For LogStream filter, we have multiple "types" of filters. Being one of: 'prefix', 'all', or 'specific'. We want to capture what type of filter is being used in VSCode. 
Conversely, there is no "type" to a logEventFilter. It is always a text pattern, thus why it is set as a boolean.

Are we able to emit the same metric twice in the same `telemetry.run()` and get two data points? Say we have `filterPattern` as the metric name as suggested, and we have two properties that have filters. Would this lock us in to only emitting one `filterPattern` value per request?

This doesn't immediately impact us since we only have one case where we are emitting a "filterType", but I could see this being problematic if we wanted to emit a logStreamFilterType AND a logEventFilterType.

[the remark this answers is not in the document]
keeganirby: If we change the type's name to just `filterType`, and we have multiple `filterType` for a given request, how would that work? 

In our case this isn't much of an issue since the only filterType we have is for LogStreams. But hypothetically if we were also emitting a `filterType` for LogEvents, how could we set both for the same request?

Also if I make this an enum, can I make the accepted values specific to CWL (`logStream_prefix`) or should it be generic (`prefix`)?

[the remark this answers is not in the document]
justinmk3: > But hypothetically if we were also emitting a filterType for LogEvents, how could we set both for the same request?

Would it be on a single metric? 

> if I make this an enum, can I make the accepted values specific to CWL (`logStream_prefix`) or should it be generic (`prefix`)?

Feel free to add any and all useful values to an enum, but starting with generic 'prefix', 'all', 'specific' seems fine. Fields exist on metrics, thus they are inherently "namespaced" by whatever metric they are attached to.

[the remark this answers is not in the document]
keeganirby: Yeah in a single metric. Again this is mainly a hypothetical, but we are currently emitting a single metric per "startLiveTail" request. If we wanted to emit a filterType for two different parameters on that request, that might be problematic.

This isn't a blocker to this PR, so i'm moving forward with renaming to `filterType` with allowedValues: ['all', 'prefix', 'specific']

PR #932, opened 2024-12-05
fix(cwl): Rename LiveTail metric definitions

keeganirby:
## Problem
Addresses feedback on https://github.com/aws/aws-toolkit-common/pull/916.

## Solution
* Uses `hasTextFilter` instead of defining a new `hasLogEventFilterPattern` property
* Defines a more generic `filterPattern` type, as an enum that supports logStream filter type strings
* renames metrics to `cloudwatchlogs_startLiveTail` and `cloudwatchlogs_stopLiveTail`

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1265,6 +1265,16 @@
             "description": "Filename extension (examples: .txt, .yml, .yaml, .asl.yaml, ...), or empty string if the filename does not contain dot (.) between two chars."
         },
         {
+            "name": "filterType",
+            "type": "string",
+            "allowedValues": [
+                "all",
+                "prefix",
+                "specific"
+            ],
+            "description": "Type of filter applied"
+        },
+        {
             "name": "findingId",
             "type": "string",
             "description": "The id of a security finding from a code scan"
@@ -1296,11 +1306,6 @@
             "name": "hasChatAuth",
             "type": "boolean",
             "description": "Whether the user has access to CodeWhisperer Chat"
-        },
-        {
-            "name": "hasLogEventFilterPattern",
-            "type": "boolean",
-            "description": "If LogEvent filter pattern is applied"
         },
         {
             "name": "hasTextFilter",
@@ -1442,11 +1447,6 @@
             "name": "locale",
             "type": "string",
             "description": "User locale. Examples: en-US, en-GB, etc."
-        },
-        {
-            "name": "logStreamFilterType",
-            "type": "string",
-            "description": "Type of LogStream filter applied to session"
         },
         {
             "name": "metricId",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -3645,6 +3645,48 @@
...
+                }
+            ]
+        },
+        {
+            "name": "cloudwatchlogs_stopLiveTail",
+            "description": "When user stops a liveTailSession",
+            "metadata": [
+                {
+                    "type": "duration",
+                    "required": true
```
justinmk3, line 3679: duration is a standard field. https://github.com/aws/aws-toolkit-common/blob/main/telemetry/telemetryformat.md#global-arguments

was there a client problem that prompted you to add this field here?

telemetry/definitions/commonDefinitions.json
```diff
@@ -3645,6 +3645,48 @@
...
+                },
+                {
+                    "type": "result"
```
justinmk3, line 3682: result is a standard field, don't need to add it here

telemetry/definitions/commonDefinitions.json
```diff
@@ -5010,48 +5052,6 @@
...
-                },
-                {
-                    "type": "logStreamFilterType",
-                    "required": false
-                },
-                {
-                    "type": "result"
-                },
-                {
-                    "type": "sessionAlreadyStarted",
```
justinmk3, line 5029: Just FYI: if "sessionAlreadyStarted" is a kind of failure, it's strongly recommended to simply throw an exception and let `telemetry.xx.run(...)` automatically set `result=Failed` and `reason=<code>` where `<code>` is the error code.  https://github.com/aws/aws-toolkit-vscode/blob/016e478399f4e4e050e5e644c98c96297930e8ad/packages/core/src/shared/errors.ts#L144

That has other benefits: all failures can be captured in dashboards/queries easily, without needing to know specific, special fields such as this one.

However, if this is only an informational field and doesn't represent a failure or cancellation, then the above doesn't apply.

telemetry/definitions/commonDefinitions.json
```diff
@@ -3645,6 +3645,48 @@
...
+                },
+                {
+                    "type": "hasTextFilter",
+                    "required": false
+                },
+                {
+                    "type": "result"
+                },
+                {
+                    "type": "sessionAlreadyStarted",
```
justinmk3, line 3664: Just FYI: if "sessionAlreadyStarted" is a kind of failure, it's strongly recommended to simply throw an exception and let `telemetry.xx.run(...)` automatically set `result=Failed` and `reason=<code>` where `<code>` is the error code.  https://github.com/aws/aws-toolkit-vscode/blob/016e478399f4e4e050e5e644c98c96297930e8ad/packages/core/src/shared/errors.ts#L144

That has other benefits: all failures can be captured in dashboards/queries easily, without needing to know specific, special fields such as this one.

However, if this is only an informational field and doesn't represent a failure or cancellation, then the above doesn't apply.

keeganirby: We don't consider this to be a failure case. This would be an informational field. The `telemetry.xx.run(...)` will not raise an exception.

telemetry/definitions/commonDefinitions.json
keeganirby: No client problem. In the PR where these were initially added it was commented on to add these fields. If they are not explicitly needed then I will remove them

PR #933, opened 2024-12-05
refactor(amazonq): remove unused metric

dhasani23:
## Problem

`codeTransformStartSrcComponents` is not used anywhere, and `result` / `reason` are a part of every metric by default, so they don't need to be manually included.

## Solution

Remove them.

## License


justinmk3 (approved):
Thank you!

PR #938, opened 2024-12-12
telemetry(amazonq): add job ID to more metrics

dhasani23:
## Problem

It can be difficult to track what happens to transformation jobs without the job ID attached to each metric.

## Solution

Add it where missing.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -3915,6 +3915,10 @@
                     "required": true
                 },
                 {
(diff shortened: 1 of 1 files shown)


justinmk3:
Just for reference, note also the `traceId` is a more generalized concept, though it of course doesn't map to a service-side job-id. https://github.com/aws/aws-toolkit-common/blob/bd5956cde9f5139b0d5e7fea82fa2f94b3a2c384/telemetry/telemetryformat.md?plain=1#L95

PR #961, opened 2025-01-21
telemetry(lsp): standardize setup metric and types

Hweinstock:
## Problem
Relevant PR: https://github.com/aws/aws-toolkit-vscode/pull/6385

## Solution
- standardize the metrics across the toolkits. 

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -5996,7 +5996,7 @@
         },
         {
             "name": "languageServer_setup",
-            "description": "LSP setup event",
+            "description": "Sets up a language server",
             "metadata": [
                 {
                     "type": "id"
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1414,16 +1414,16 @@
             "description": "Language used for the project."
         },
         {
-            "name": "languageServerSetupStage",
-            "type": "string",
-            "allowedValues": ["getManifest", "getServer", "validate", "launch", "final"],
-            "description": "The stage of the LSP setup process"
-        },
-        {
             "name": "languageServerResourceLocation",
             "type": "string",
             "allowedValues": ["cache", "remote", "fallback", "override", "unknown"],
             "description": "The location of the LSP resource"
+        },
+        {
+            "name": "languageServerSetupStage",
+            "type": "string",
+            "allowedValues": ["getManifest", "getServer", "validate", "launch", "final"],
+            "description": "The stage of the LSP setup process"
         },
         {
             "name": "loadFileTime",
@@ -5993,12 +5993,12 @@
             "description": "LSP setup event",
             "metadata": [
                 {
-                    "type": "languageServerSetupStage",
-                    "required": true
-                },
-                {
                     "type": "languageServerResourceLocation",
                     "required": false
+                },
+                {
+                    "type": "languageServerSetupStage",
+                    "required": true
                 },
                 {
                     "type": "manifestVersion",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1426,11 +1426,6 @@
             "description": "The stage of the LSP setup process"
         },
         {
-            "name": "languageServerVersion",
-            "type": "string",
-            "description": "The version of the language server"
-        },
-        {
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1413,6 +1413,18 @@
             "type": "string",
             "description": "Language used for the project."
         },
+        {
+            "name": "languageServerResourceLocation",
```
justinmk3, line 1417: Why do we continually type out "language server" everywhere? It is less consistent and harder to guess (when grepping) than just typing "lsp". It's also potentially "wrong" some of these metrics will be related to the "client", not the "server". 

```suggestion
            "name": "lspResourceLocation",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -1423,6 +1435,11 @@
             "type": "string",
             "description": "User locale. Examples: en-US, en-GB, etc."
         },
+        {
+            "name": "manifestVersion",
```
justinmk3, line 1439: we also have `buildSystemVersion`. can we name this `version` so that we don't need specific "fooVersion" fields for each topic

telemetry/definitions/commonDefinitions.json
```diff
@@ -5971,6 +5988,24 @@
                 }
             ]
         },
+        {
+            "name": "languageServer_setup",
+            "description": "LSP setup event",
```
justinmk3, line 5993: the namespace can be "lsp". then each mertic can say if it's clien-specific or server-specific, assuming that is something the metric wants to call out.

```suggestion
            "name": "lsp_setupServer",
            "description": "Started an LSP server",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -5971,6 +5988,24 @@
...
+                {
+                    "type": "languageServerResourceLocation",
+                    "required": false
+                },
+                {
+                    "type": "languageServerSetupStage",
+                    "required": true
+                },
+                {
+                    "type": "manifestVersion",
```
justinmk3, line 6004: `manifestVersion` seems like a necessary "specific" metric. Do we also need a `version` field which here can be used to indicate the lsp server version? Or could we just name this field `version`.

telemetry/definitions/commonDefinitions.json
Hweinstock: I agree, but it does look like both eclipse and VS use the `languageServer` prefix, so we'd have to come to agreement on this transition.

telemetry/definitions/commonDefinitions.json
```diff
@@ -5971,6 +5994,36 @@
                 }
             ]
         },
+        {
+            "name": "languageServer_setup",
+            "description": "LSP setup event",
+            "metadata": [
+                {
+                    "type": "id",
+                    "required": true
```
shruti0085, line 6003: metadata fields are by default true. Only those meant to be optional need the specification of false.

telemetry/definitions/commonDefinitions.json
```diff
@@ -5971,6 +5994,36 @@
                 }
             ]
         },
+        {
+            "name": "languageServer_setup",
+            "description": "LSP setup event",
```
shruti0085, line 5999:
```suggestion
            "description": "Sets up a language server",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -5971,6 +5994,36 @@
...
+            "metadata": [
```
shruti0085, line 6000: This is a passive event because users do not take an actual action for this to occur, please indicate the same. In addition fields of duration and result is missing from this shape

telemetry/definitions/commonDefinitions.json
dhruvigajjar: I think VS has implemented it as "manifestSchemaVersion" because there was need of multiple version fields. such as languageServerVersion & Manifest's version.

telemetry/definitions/commonDefinitions.json
justinmk3: > but we don't have support for `result`, so this field needs to be added to the metadata.

Can that be prioritized? Adding `result` explicitly to every metric confuses the messaging for partner teams.

justinmk3 (approved):

PR #963, opened 2025-01-22
telemetry(amazonq): add fields for code fix events

ctlai95:
## Problem

Need to understand:
- How many users are using fix feature from /review (not auto-review)
- How many fixes were generated/applied for each scanJob
- How many code fix requests were unable to generate a fix

## Solution

- Add flag `autoDetected` to be able to filter out auto-review findings
- Add flag `fixGenerated` to be able to see fixes that generated no fixes
- Add `codewhispererCodeScanJobId` to associate a fix action with a specific scan that produced the finding

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -159,6 +159,11 @@
             "description": "The type of auth flow used for signing in"
         },
         {
+            "name": "autoDetected",
+            "type": "boolean",
+            "description": "Whether the code issue was detected by auto-review"
+        },
+        {
             "name": "awsAccount",
             "type": "string",
             "description": "AWS account ID associated with a metric.\n- \"n/a\" if credentials are not available.\n- \"not-set\" if credentials are not selected.\n- \"invalid\" if account ID cannot be obtained."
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1287,6 +1292,11 @@
             "type": "int",
             "description": "Number of findings discovered after executing IAM Policy Checks"
         },
+        {
+            "name": "fixGenerated",
```
justinmk3, line 1296: should this share a common prefix with the exsting `codeFix..` field? helps discoverability, assuming they are somewhat related.

https://github.com/aws/aws-toolkit-common/blob/ab466029f241193ab93a8abb3a5ca8dde5f56c94/telemetry/definitions/commonDefinitions.json#L286

```suggestion
            "name": "codeFixGenerated",
```

telemetry/definitions/commonDefinitions.json
```diff
@@ -1287,6 +1292,11 @@
...
+            "type": "boolean",
+            "description": "Whether a code fix request was able to generate a fix"
```
justinmk3, line 1298: How is this different from `includesFix` ? https://github.com/aws/aws-toolkit-common/commit/24ebf0638f09fc9ac1a0d7343ef8d632bbcde28c

justinmk3 (approved):

telemetry/definitions/commonDefinitions.json
ctlai95: Good catch, I will reuse this one

PR #977, opened 2025-02-19
telemetry: add `source` to common fields

nkomonen-amazon:
## Problem:

The `source` value has become a common pattern in our telemetry which answers "who" caused the metric event. For example if we had a function() to open a menu, there could be many different entrypoints that could call it.

Right now we have to add `source` to each metric that uses it, but this usage has increased for a lot of metrics.

## Solution:

- `source` will be a part of all metrics as an optional field.
- This has only been implemented for VSC.

## TODO

The other IDEs should consider following this pattern if not already.
**If all IDEs implement this then we can remove the explicit requirements of `source` in `commonDefinitions.json`**

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1782,7 +1782,7 @@
         {
             "name": "source",
             "type": "string",
-            "description": "The source of the operation"
+            "description": "The source of the operation. This answers 'who' caused/triggered the operation. Example: did an Auth signout happen because of some expiration or since the user explicitly clicked the signout button."
         },
         {
             "name": "sourceFacetType",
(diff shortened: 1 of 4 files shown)


awschristou:
Can you add this to the C# generator too?

telemetry/definitions/commonDefinitions.json
```diff
@@ -1782,7 +1782,7 @@
         {
             "name": "source",
             "type": "string",
-            "description": "The source of the operation"
+            "description": "The source of the operation. This answers 'who' caused/triggered the operation. Example: did an Auth signout happen because of some expiration or since the user explicitly clicked the signout button."
```
justinmk3, line 1785: Nice doc enhancement 👍

justinmk3 (approved):

jpinkney-aws (approved):

PR #984, opened 2025-02-26
doc: update metric naming spec

nkomonen-amazon:
## Problem

We do not have a clear defined format for metric names.

## Solution

Update the doc that defines how to name a metric.

- This will NOT retroactively apply to all existing metrics since this will require significant work refactoring that we do not want to do at the moment.

- The benefits of this new structure `{namespace}_{noun}{Verb}` are that it will group related metrics in the definitions file since they are alphabetically ordered. Also when getting IDE completions all relevant options will be grouped together. This should help for searchability/reusability.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #985, opened 2025-02-27
refactor: rename existing toolkit module metric names to new format 

nkomonen-amazon:
## Problem:

We have a new telemetry metric name format defined in our telemetry
readme. The current module metrics do not follow this spec.

## Solution:

- Rename the existing metrics to match #984 spec. Eg: `toolkit_openModule` -> `toolkit_willOpenModule`
- Reorder the renamed metric alphabetically
- Note we are not doing this for everything, but for some select metrics.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -7156,7 +7156,7 @@
         },
         {
             "name": "toolkit_moduleClose",
-            "description": "The user closed 'something' (specified by 'module'). Examples: a view, feature, resource, ...",
+            "description": "The module has closed/terminated. Use this when the module is done being used.",
             "metadata": [
                 {
                     "type": "module",
@@ -7166,8 +7166,8 @@
             "passive": true
         },
         {
-            "name": "toolkit_moduleInit",
-            "description": "The Toolkit has completed initialization for the specified module.",
+            "name": "toolkit_moduleLoad",
+            "description": "The module has loaded, i.e it has rendered/resolved/finished. Use this after `toolkit_moduleOpen` to indicate success.",
             "metadata": [
                 {
                     "type": "attempts",
@@ -7194,7 +7194,7 @@
         },
         {
             "name": "toolkit_moduleOpen",
-            "description": "User opened 'something' (specified by 'module'). Examples: a view, feature, resource, ...",
+            "description": "The opening of a module **started**. This does not imply it successfully finished, use `toolkit_moduleLoad` for that. The open/load separation exists because it may be difficult to capture then entire process in a single metric. A 'module' is a general term that represents things like: a view, feature, resource, ... ",
             "metadata": [
                 {
                     "type": "module",
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -7060,6 +7060,44 @@
             ]
         },
         {
+            "name": "toolkit_didCloseModule",
+            "description": "The module has closed/terminated. Use this when the module is done being used.",
+            "metadata": [
+                {
+                    "type": "module",
+                    "required": true
+                }
+            ],
+            "passive": true
+        },
+        {
+            "name": "toolkit_didLoadModule",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -7136,65 +7125,76 @@
...
                 }
-            ]
+            ],
+            "passive": true
         },
         {
-            "name": "toolkit_openModule",
-            "description": "User opened 'something' (specified by 'module'). Examples: a view, feature, resource, ...",
+            "name": "toolkit_moduleOpen",
+            "description": "The opening of a module **started**. This does not imply it successfully finished, use `toolkit_moduleLoad` for that. The open/load separation exists because it may be difficult to capture then entire process in a single metric. A 'module' is a general term that represents things like: a view, feature, resource, ... ",
```
awschristou, line 7197: This is going to lead to confusion: "Open" means "open started" but "Load" means "Load completed". Should we use verb tense as part of our naming convention, so that we have "moduleOpen" and "moduleLoaded" ?  Also, is there merit in renaming "moduleOpen" to "moduleLoad" ?

nkomonen-amazon: Great points, I chatted with some people in standup and we considered:
- "toolkit_moduleOpening"
- "toolkit_moduleLoaded"

Do you like this?

justinmk3: > "Open" means "open started" but "Load" means "Load completed"

This is new scope that was not previously considered in these metric names. Usually, the description defines the semantics.

Using English verb endings like "ed" and "ing" doesn't always work, some of these names aren't necessarily grammatical English (they are technical nonces).

If we really want to indicate "pre" and "post" stages of metrics in their names (instead of just the descrpitions), it will be less ambiguous to use the `did` and `will` prefixes (lsp examples: [didRename](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didRenameFiles), [willDelete](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willDeleteFiles)).

But are we sure this is worth the trouble? The sequence isn't always well-defined, sometimes a "will" becomes a "did" later because of code changes, but the name will then be misleading. Instead the description can clear any confusion.

telemetry/definitions/commonDefinitions.json
```diff
@@ -7136,65 +7125,76 @@
...
                 {
-                    "type": "version",
-                    "required": false
+                    "type": "source",
+                    "required": true
+                }
+            ]
+        },
+        {
+            "name": "toolkit_moduleClose",
```
justinmk3, line 7158: The `namespace-noun-verb` ordering is something I missed in the updated "spec". This is contrary to almost all of our existing metrics, isn't it? `auth_signInPageOpened`, `auth_switchRoles`, `aws_refreshExplorer`.

Can we just define the spec as `namespace-verb-noun`, instead? That will avoid a lot of churn and/or conflicting legacy names.

nkomonen-amazon: `auth_signInPageOpened` in this case does conform to the `namespace-noun-verb` spec. So I guess this was a bit of an example that we didn't have well defined standards already.

I mentioned the reason for doing`namespace-noun-verb` here: https://github.com/aws/aws-toolkit-common/pull/984/files#diff-809396fc93ef2f298c19a75db76447c555a86af06581d65d65808533cbc10769R33

But I'm open to discussion so we can define some standard for the future

nkomonen-amazon: Summary of discussion w/ Justin + Chris:

- We will update spec to more closely match our existing metrics
  - `{namespace}_{did|will| }{verb}{noun}`
  - Examples:
    - `toolkit_openModule`
    - `toolkit_willOpenModule`
    - `toolkit_didLoadModule`

- IF we didn't have these legacy metrics we would have considered a different scheme
  - Example: `{namespace}_{noun}_{did|will| }{verb}`
  - But to reduce confusion since all our existing metrics tend to follow {verb}{noun}

nkomonen-amazon:
@awschristou @justinmk3 can one of you please give a final look so we can merge

awschristou (approved):

PR #1005, opened 2025-03-21
telemetry(amazonq): add id2 metric

zuoyaofu:
## Solution

Some users are reporting S3 upload failures but we don't have the id2 that S3 team requires to investigate further.
Emit amazonq_createUpload metric with id2, and add logs for these ids so customers can use too.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1550,11 +1550,6 @@
             "description": "Metric-defined identifier"
         },
         {
-            "name": "id2",
-            "type": "string",
-            "description": "x-amz-id-2 for S3 Upload Debugging"
-        },
-        {
             "name": "includesFix",
             "type": "boolean",
             "description": "Whether the security issue includes a suggested fix"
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1549,6 +1549,11 @@
             "type": "string",
             "description": "Metric-defined identifier"
         },
+        {
+            "name": "id2",
```
awschristou, line 1553: This name is too generic to be meaningful. "eg: Whats id2 is it more special than the `id` field?"

If it is s3 specific, would it make sense to name it `s3_id2`? Could it be generalized to something like `responseHeader`?

telemetry/definitions/commonDefinitions.json
```diff
@@ -1549,6 +1549,11 @@
...
+            "type": "string",
+            "description": "x-amz-id-2 for S3 Upload Debugging"
```
awschristou, line 1555: Its great to see an example, but please add a proper description to explain the field.

telemetry/definitions/commonDefinitions.json
zuoyaofu: Will update! Thanks for pointing that out.

awschristou (approved):
I'm approving the change to unblock some priority work. There's a good chance this new field will be revisited for a potential rename.

PR #1013, opened 2025-04-15
telemetry(amazonq): agentic loop

chungjac:
## Problem
- want to figure out if a message is apart of an agentic loop or not

## Solution
- add field `AgenticChatWithToolUse `

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1269,6 +1269,11 @@
             "description": "Truncated length of workspace context"
         },
         {
+            "name": "cwsprIsAgenticLoop",
+            "type": "boolean",
+            "description": "If the message is apart of an agentic loop (LLM asks to run a tool and receives results back)"
+        },
+        {
             "name": "databaseCredentials",
             "type": "string",
             "allowedValues": [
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1268,6 +1268,11 @@
             "type": "int",
             "description": "Truncated length of workspace context"
         },
+        {
+            "name": "cwsprIsAgenticLoop",
```
justinmk3, line 1272: booleans should be avoided. why can't the existing fields be used? 

`cwsprAgenticChatInteractionType` https://github.com/aws/aws-toolkit-common/blob/3ee19949c3f5e26ecb5caf04986d7d9f38e4fa5f/telemetry/definitions/commonDefinitions.json#L976

https://github.com/aws/aws-toolkit-common/blob/3ee19949c3f5e26ecb5caf04986d7d9f38e4fa5f/telemetry/definitions/commonDefinitions.json#L1018-L1024

If a new field is still needed, should it be a field like `cwsprMessageType` instead of adding booleans for every new permutation ?

chungjac: Fixed!

PR #1025, opened 2025-05-14
telemetry(amazonq): adding enabled and perfE2ELatency to agentic chat metrics

laileni-aws:
## Problem
- Missing metric field to separate agentic coding mode.
- Need to add a new field for amazonq_toolUseSuggested metric to calculate the latency of tools.
## Solution
- Fixed above issues.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2455,6 +2455,10 @@
                     "required": false
                 },
                 {
+                    "type": "enabled",
+                    "required": false
+                },
+                {
                     "type": "languageServerVersion",
                     "required": false
                 },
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -3047,9 +3055,17 @@
                 {
                     "type": "cwsprToolUseId"
                 },
+                {
+                    "type": "enabled",
```
samgst-amazon, line 3059: is this the name that will appear in metrics? why not just call it pairProgrammingEnabled or something

laileni-aws: I am trying to reuse the existing field which is available! instead of creating new fields.

samgst-amazon (approved):

PR #1051, opened 2025-06-13
telemetry: add metrics for pinned context

avi-alpert:
Add telemetry for pinned context

### Pinned Context Counts
- `cwsprChatPinnedCodeContextCount`: Tracks the number of pinned code symbols added to context
- `cwsprChatPinnedFileContextCount`: Tracks the number of pinned files added to context
- `cwsprChatPinnedFolderContextCount`: Tracks the number of pinned folders added to context
- `cwsprChatPinnedPromptContextCount`: Tracks the number of pinned saved prompts added to context

### Rule Context Tracking
- `cwsprChatTotalRuleContextCount`: Tracks the total number of workspace rules (both active and inactive)
- Updated description for `cwsprChatRuleContextCount` to clarify it only includes active rules

All new metrics have been added to `amazonq_addMessage` and are marked as optional fields.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -1161,6 +1161,26 @@
             "description": "Unique identifier for each message in an conversation"
         },
         {
+            "name": "cwsprChatPinnedCodeContextCount",
+            "type": "int",
+            "description": "Number of pinned code symbols added to context"
+        },
+        {
+            "name": "cwsprChatPinnedFileContextCount",
+            "type": "int",
+            "description": "Number of pinned files added to context"
+        },
+        {
+            "name": "cwsprChatPinnedFolderContextCount",
+            "type": "int",
+            "description": "Number of pinned folders added to context"
+        },
+        {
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -1160,6 +1160,26 @@
             "type": "string",
             "description": "Unique identifier for each message in an conversation"
         },
+        {
```
laileni-aws, line 1163: Can we make these fields more generic like `codeCount`, `fileCount` etc to make it reusable?

avi-alpert: The names of the types im adding use the same naming convention as the existing context count types: https://github.com/aws/aws-toolkit-common/blob/416c47973f75917f1e0ae7ba0b67df97848ea878/telemetry/definitions/commonDefinitions.json#L1174

except with the added prefix `Pinned`

this way we can keep track which context items were added in the prompt vs which were pinned

jngyuamz (approved):

PR #1070, opened 2025-07-23
telemetry: add metrics for iam credentials authentication

yuxianrz:
## Metrics Modified:
`auth_addConnection`: add `credentialType` type to record distinguish IAM and other login method.

## License

### DIFF
--- a/telemetry/definitions/commonDefinitions.json
+++ b/telemetry/definitions/commonDefinitions.json
@@ -2134,6 +2134,16 @@
             "description": "The state of the setting being changed to. This should not be recorded for free-form settings like file-system paths. Instead, stick to things like flags, numbers, and enums."
         },
         {
+            "name": "signInOption",
+            "type": "string",
+            "allowedValues": [
+                "Company account",
+                "IAM Credentials",
+                "Other"
+            ],
+            "description": "The sign in option for AmazonQ extension."
+        },
+        {
             "name": "source",
             "type": "string",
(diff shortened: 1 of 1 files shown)


telemetry/definitions/commonDefinitions.json
```diff
@@ -3988,6 +3988,24 @@
                 }
             ]
         },
+        {
```
laileni-aws, line 3991: Can we combine these two metric names into a single metric?
// Keep this generic if required we can add a type: IAM || BuilderId etc
Example:

```
{
            "name": "auth",
            "description": "Emitted for authentication",
            "metadata": [
                {
                    "type": "credentialType",
                },
                {
                    "type": "reason",
                },
                {
                     "type": "reasonDesc",
                 },
// Add any other fields here....
            ]
        },

Comment (yuxianrz) [reply telemetry/definitions/commonDefinitions.json:4001]: They are capturing two quite different behaviors so we do not want to combine these two. But I made a new commit and changed these two metrics to capture option click and login failure for all credentialType. In this way, they are more generic as are not just used for IAM authflow.

File: telemetry/definitions/commonDefinitions.json [hunk 2]
```diff
@@ -3988,6 +3998,30 @@
...
+            "name": "auth_loginFailure",
+            "description": "Emitted when User submit IAM/SSO form but receive error",
+            "metadata": [
+                {
+                    "type": "credentialType",
+                    "required": true
+                },
+                {
+                    "type": "reason",
+                    "required": true
```
laileni-aws, line 4011: By default required is true

laileni-aws (approved):
LGTM

